PR #10, opened 2019-05-17
Convert DAS to use MQTT protocol!

khanguslee:
Just need to update the repository..

Hopefully we do smaller PRs from now on!


Raspi/DAS.js
```diff
@@ -0,0 +1,175 @@
...
+      mqttClient.publish('stop');
+    }
+
+    if (IS_RECORDING) {
+      const currentTime = Math.floor(Date.now());
+      const outputData = `${data}&filename=${FILENAME}&time=${(currentTime - INITIAL_TIME)}&power=${POWER}&cadence=${CADENCE}`;
+      mqttClient.publish('data', outputData);
+    }
+  });
+});
```
harsilspatel: One

harsilspatel, line 175: 1. Definitely need linters.
2. Should break down file into multiple modules

Raspi/Test/mqtt_test.py
```diff
@@ -0,0 +1,66 @@
...
+broker_address = args.host
+client = mqtt.Client()
+
+client.on_connect = on_connect
+client.on_message = on_message
+
+client.connect(broker_address)
+start_publishing(client)
+
+client.loop_forever()
```
harsilspatel, line 66: We should definitely include more tests.

Raspi/DAS.js
khanguslee: Definitely the next steps to do. But will need to update master branch with what we have at the moment. Otherwise the PR's get too big (like this one).

harsilspatel (approved):
1. We definitely need linters.
2. Should consider getting rid of the .lock files.
3. We better setup travis.
4. Please commit frequently :)

Raspi/Test/mqtt_test.py
khanguslee: I would love to have some node.js endpoint tests + unit tests. Added onto backlog.

PR #11, opened 2019-05-17
Update dependencies

harsilspatel:


harsilspatel:
It seems that the latest version of mqtt (v2.18.8) is dependent on v5.6.0 of mqtt-packet, which github identifies as malicious.

khanguslee (approved):
looks good to me.

PR #13, opened 2019-07-12
Replace gps_location field with gps_lat, gps_long and gps_alt

hallgchris:
Depends on Monash-Human-Power/MHP-DAS-Web-Server#15

### DIFF
--- a/Raspi/Test/mqtt_test.py
+++ b/Raspi/Test/mqtt_test.py
@@ -22,7 +22,7 @@ def start_publishing(client):
         total_time = round(current_time - start_time, 2)
 
         # Generate data message
-        data = "gps=1&gps_location=00&gps_course=00&"
+        data = "gps=1&gps_lat=00&gps_long=00&gps_alt=00&gps_course=00&"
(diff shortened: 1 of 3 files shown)


harsilspatel (approved):

harsilspatel:
good work @hallgchris 
i think we should also setup pylint at some point :)

hallgchris:
> i think we should also setup pylint at some point :)

I actually use pylint in my editor and that would give the excuse to get rid of all the editor warnings in everyone's code :joy: I vote yes

khanguslee:
> good work @hallgchris
> i think we should also setup pylint at some point :)

Sounds good to me. Can someone make a PR hehehe

khanguslee (approved):
I'm gonna approve this. However, I will need to check this manually with the real DAS before we can confirm that it works.

hallgchris:
[Log longitude, latitude and altitude in separate columns](https://trello.com/c/1f5gCuWt/2-log-longitude-latitude-and-altitude-in-separate-columns)

PR #14, opened 2019-07-12
Update mqtt_test.py to replay a previously recorded csv file over MQTT

hallgchris:


hallgchris:
[MQTT DAS test script to output actual race data](https://trello.com/c/KfoBGEFh/36-mqtt-das-test-script-to-output-actual-race-data)

hallgchris:
[MQTT DAS test script to output actual race data](https://trello.com/c/KfoBGEFh/36-mqtt-das-test-script-to-output-actual-race-data)

harsilspatel (approved):
looks good to me!

harsilspatel:
do we also want to write some test cases to ensure that the newly added features work or would that be an overkill? 🤔

khanguslee:
> do we also want to write some test cases to ensure that the newly added features work or would that be an overkill? 🤔

Not sure how you would test arduino code. 

Also doesn't make sense to make a test for your test? But we certainly do need tests in the other files!

khanguslee (approved):
This is exactly what I always wanted. Good job @hallgchris ! 

Only thing I have to say is to make sure you branch off for different features you are working on. Good catch on passing args into the functions!

PR #16, opened 2019-07-16
mqtt_test.py: Publish filename when replaying CSVs

hallgchris:
Forgot to define the filename when replaying a CSV, so it showed as undefined. It now will display the name of the file being replayed.


khanguslee (approved):

harsilspatel (approved):
looks good to me :)

PR #18, opened 2019-08-22
Cleanup repository

khanguslee:
Also DAS scripts now print out a short decription of what program it is. This allows for a team member to view what script is currently loaded onto the Teensy.

Removed some trailing whitespaces too.


.gitignore
```diff
@@ -3,4 +3,5 @@
 .vscode
 
 node_modules/
-.eslintrc.*
\ No newline at end of file
+.eslintrc.*
```
hallgchris: Why don't we want .eslintrc? I'd have thought it should be here for CI and so we're all following the same standards

hallgchris (approved):
LGTM

.gitignore
khanguslee: I swear this was added in automatically LMAO. I'll remove this.

harsilspatel (approved):
yeah good work @khanguslee, merging it ^_^

PR #28, opened 2019-10-07
Add serial_test.py

hallgchris:
Basically mqtt_test.py but imitates the serial data coming from the teensy. To test:

`python serial_test.py -f data_173.csv -j 1500` (or whatever file you want)

Then in another console,

`node DAS.js -a -p [slave port]` (slave port is printed by the previous statement)

Then press enter in the first console window. You can confirm it's working if you see data on `mosquitto_sub -t data` or on the DAShboard.

Tested on linux, should work on macOS, don't think it will work on windows.


khanguslee:
Will need to test this out. But on a quick glance, this is looking good. Only gripe I have is to rename `test_data_generator.py` to something else as it may run as a test script from pytest or similar due to the `test_*.py` filename convention.

harsilspatel (dismissed):
looks great, nice work bikie! @hallgchris ^_^

Raspi/Test/serial_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+def on_connect(client, userdata, flags, rc):
+    print("Connected to MQTT broker with result code " + str(rc))
+
+def run(args):
+    # Create pseudo-terminal pair
+    master, slave = os.openpty()
+
+    # Show ports
+    print("Master serial port:", os.ttyname(master))
```
khanguslee, line 30: This is a Unix only method. https://docs.python.org/3/library/os.html#os.ttyname
We should probably check for operating system. If the user is trying to run it on Windows, let's just return a error message stating this won't work!

khanguslee, line 30: Also I get an error here.
Screen Shot 2019-10-08 at 9 39 41 pm

I get a master value of 3 and a slave value of 4.

khanguslee (dismissed):
Have an issue with running the script. Might be good to update the `README.md` on how to run this script.

Raspi/Test/serial_test.py
hallgchris: Done

hallgchris: Would you be able to have another go at this? If it still doesn't work I might need to get my hands on a mac to test on

khanguslee: I still get the same thing :(
Screen Shot 2019-12-15 at 12 05 47 am

Raspi/Test/serial_test.py
```diff
@@ -0,0 +1,55 @@
...
+    print("Connected to MQTT broker with result code " + str(rc))
+
+def run(args):
+    # Check that we are running on a unix system.
+    # os.openpty and os.ttyname are only available on unix (NOT windows)
+    if os.name != "posix":
+        raise Exception("Serial test only works on Unix systems. If you are on Windows, perhaps try using WSL.")
+
+    # Create pseudo-terminal pair
+    master, slave = os.openpty()
```
emilytrau, line 27: Maybe try `pty.openpty` instead? It has some extra fallbacks in case it fails

hallgchris: @khanguslee can you try this? Should be able to replace `os.openpty()` with `pty.openpty()` and run

khanguslee: Unfortunately I get the same problem :(

khanguslee:
If this is working on linux, happy to just accept this but with an OS check that only allows the test script to work on linux only. We'll worry about getting this to work for other OS' in another task :)

Thoughts?

hallgchris:
> If this is working on linux, happy to just accept this but with an OS check that only allows the test script to work on linux only. We'll worry about getting this to work for other OS' in another task :)
> 
> Thoughts?

Happy with that for now. I'm not really sure if I'll be able to work out a fix without having a mac to experiment on, so otherwise the PR will be sitting here indefinitely

hallgchris:
Pretty sure everything's fine but gotta get reviewed again to merge :sweat_smile:

emilytrau (dismissed):
Looks good to me

Raspi/Test/serial_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+from das_data_generator import send_csv_data, send_fake_data
+
+# Arguments
+parser = argparse.ArgumentParser(description='DAS serial test script', add_help=True)
+parser.add_argument('-t', '--time', action='store', type=int, default=1, help="length of time to send data")
+parser.add_argument('-r', '--rate', action='store', type=float, default=0.5, help="rate of data in seconds")
+parser.add_argument('--host', action='store', type=str, default="localhost", help="address of the MQTT broker")
+parser.add_argument('-f', '--file', action='store', type=str, help="The csv file to replay. If not specified, makes up data.")
+parser.add_argument('-j', '--jump', action='store', type=int, default=0, help="Starts replaying from a specified time (in seconds)")
```
harsilspatel, line 14: i just love how the arguments also have shorthands!

harsilspatel (dismissed):
as always, great work chris! :D

hallgchris:
Sorry guys need another review :joy:

emilytrau (approved):
Looks good 👍

hallgchris:
Finally lol

PR #32, opened 2020-01-12
Output mock GPS from test script

emilytrau:
## Description

Modified the test data generator to output reasonable GPS locations.

## Screenshots

image

## Steps to Test

```
cd Raspi/Test
python mqtt_test.py -t 10
```

### DIFF
--- a/Raspi/Test/das_data_generator.py
+++ b/Raspi/Test/das_data_generator.py
@@ -25,8 +25,8 @@ def send_fake_data(send_data_func, duration, rate, immitate_teensy=False):
         data += "&gps_lat=" + str(gps_lat)
         data += "&gps_long=" + str(gps_long)
         # Move up to 5 metres per second in both directions
-        gps_lat += random.randint(0, 50) * 0.1 * meter_lat * rate
-        gps_long += random.randint(0, 50) * 0.1 * meter_long * rate
+        gps_lat += random.randint(0, 50) * 0.1 * metre_lat * rate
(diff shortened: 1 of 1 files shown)


Raspi/Test/das_data_generator.py
```diff
@@ -9,14 +9,22 @@ def send_fake_data(send_data_func, duration, rate, immitate_teensy=False):
...
     while True:
         current_time = round(time.time(), 2)
         total_time = round(current_time - start_time, 2)
 
         # Generate data message
-        data = "gps=1&gps_lat=00&gps_long=00&gps_alt=00&gps_course=00&"
+        data = "gps=1"
+        data += "&gps_lat=" + str(gps_lat)
+        data += "&gps_long=" + str(gps_long)
+        gps_lat += random.randint(0, 20) * 0.00001
```
khanguslee: I would leave a comment stating why you chose those values. (eg. Random distance up to x metres away etc.)

khanguslee (changes requested):
Good idea! This will probably not cause the dashboard to freak out if we run this script now! Just a small change!

Raspi/Test/das_data_generator.py
emilytrau: I added some comments and changed the code to do movement based on metres instead of an arbitrary degree.

khanguslee:
Definitely works as intended! Very cool! 

Screen Shot 2020-01-12 at 3 35 40 pm

khanguslee (approved):
Looks good to me! Feel free to create a card on software board with what you did here!

PR #33, opened 2020-01-12
Fix errors when replaying csv data

emilytrau:
## Description

- Errors occurred when replayed data has the GPS turned off
- Time and filename data was incorrectly formatted

## Steps to Test

```
cd Raspi/Test
python mqtt_test.py -f data_173.csv
```


hallgchris (approved):
Looks great, thanks for all your hard work Angus!

PR #34, opened 2020-01-12
Add ability to speed up CSV replay

emilytrau:
## Description

Replay files can be very large and can take a long time to replay in real time. It would be beneficial for testing if we could speed it up.

## Steps to Test

```
cd Raspi/Test
# Speed up by factor of 5
python mqtt_test.py -f data_173.csv -s 5
```


khanguslee (approved):
Good idea. LGTM!

PR #35, opened 2020-01-21
Fix incorrect MQTT topics and behaviour for mock module script

Blake-Haydon:
## Description

This fixes the mock script to properly mimic the behavior of the real modules. 

- Topic names are now finally correct
- Fake battery readout only gets sent every 5 minutes instead of 5 times a second
- When started the start channel will be sent a fake file name that comprises of the sensor name and time
- When ending the stop channel will be sent blank data

## Screenshots

Screen Shot 2020-01-21 at 5 24 42 pm

The battery data only gets sent at the start once instead of each time with the data.

## Steps to Test

For 5 seconds of data (mosquitto has to be running locally)
python mock_wireless_module.py -t 5

### DIFF
--- a/Raspi/Test/mock_wireless_module.py
+++ b/Raspi/Test/mock_wireless_module.py
@@ -175,9 +175,7 @@ def publish(client, topic, data):
     print(topic, "--> ", json_data)
 
 
-def start_publishing(client, args):
-    print("publishing started...")
-
+def start_modules(args):
     # Initiate the correct modules when starting
     if args.id is None:
         for i in range(1, 4):
@@ -186,7 +184,7 @@ def start_publishing(client, args):
                 + "_" + str(round(time.time()))
                 + ".csv"
             })
-        print('started module 1\nstarted module 2\nstarted module 3')
+        print('\nstarted module 1\nstarted module 2\nstarted module 3')
 
     else:
         publish(client, "/v3/wireless-module/" + str(args.id) + "/start", {
@@ -197,12 +195,18 @@ def start_publishing(client, args):
         print('started module ' + str(args.id))
 
 
-    send_fake_data(client, args.time, args.rate, args.id)
-
+def stop_modules(args):
     for i in range(1, 4):
         publish(client, "/v3/wireless-module/" + str(i) + "/stop", {})
-    print('stopped module 1\nstopped module 2\nstopped module 3')
+    print('\nstopped module 1\nstopped module 2\nstopped module 3')
 
+
+def start_publishing(client, args):
+
+    print("\npublishing started...")
+    start_modules(args)
+    send_fake_data(client, args.time, args.rate, args.id)
+    stop_modules(args)
     print("\npublishing finished")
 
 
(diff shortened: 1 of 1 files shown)


Raspi/Test/mock_wireless_module.py
```diff
@@ -5,24 +5,24 @@
...
+    description='MQTT wireless module test script that sends fake data',
     add_help=True)
 parser.add_argument(
     '-t', '--time', action='store', type=int,
     default=float('Inf'), help="""Length of time to send data in seconds
     (duration). If nothing is selected it will continuously send out data.""")
 parser.add_argument(
-    '-r', '--rate', action='store', type=float, default=1,
-    help="""Rate of data in data sent per second. Default is 1 data pack sent
+    '-r', '--rate', action='store', type=float, default=5,
```
khanguslee: Any reason you set the default to 5?

Raspi/Test/mock_wireless_module.py
```diff
@@ -166,10 +170,44 @@ def publish(client, topic, data):
     print(topic, "--> ", json_data)
 
 
+def start_modules(args):
+    """ Send the a fake filename on the start channel to start the appropriate
+    module"""
+
+    if args.id is None:
+        for i in range(1, 4):
```
khanguslee, line 181: This should be made to take in N number of modules. Not just 3. Perhaps create another argument where we can pass in a number to create N number of modules. This will ignore/overwrite the id argument.

khanguslee (changes requested):
I would have split this PR into two separate PRs. One for fixing the MQTT topics and one for the behaviour. Notice how your PR message has an 'and' in the title? Would make for a smaller PR + more commits for you ;)

Raspi/Test/mock_wireless_module.py
Blake-Haydon: Because on this slack thread I learned the rate was currently 9hz but was going to be changed to 5hz so I changed it. Thought it represented the read data more closely.

https://monashhumanpower.slack.com/archives/CCD8JK739/p1579558297007000

Blake-Haydon: If the data rate changes its pretty easy to change back.

Raspi/Test/mock_wireless_module.py
Blake-Haydon: If N number of modules are selected should this just make up what sensors go on each module. Because at the current moment the sensors are hard coded for each module to represent the actual sensors.

Raspi/Test/mock_wireless_module.py
khanguslee: Ohhhhh, good initiative! But I think we should leave it at 1Hz for now as it's a lot easier to keep track (when testing). If we do end up using 5Hz all the time, then we'll change the default. Sorry for the confusion!

Raspi/Test/mock_wireless_module.py
khanguslee: Good question. Ignore me for now. But we should create another card that allows the user to select number of modules + what module to include. Perhaps the user should be allowed to input a config file that define module ID and sensors on that module?

khanguslee: Create a TODO on the for loops that define the above behaviour I described.

emilytrau (approved):
Looks good. Great work :+1:

hallgchris (approved):
Looks good to me! Sorry I've taken forever

khanguslee (approved):
LGTM! Your commit messages are on points btw!

PR #36, opened 2020-01-27
Wireless Module Logger

Blake-Haydon:
## Description
This pull request adds code to be able to log the data received from the wireless modules. It listens on localhost and records data when it is sent out on the MQTT data and battery topics. It waits for the start topic on MQTT to start.

`/v3/wireless-module/<id>/start`

And continues recording until the stop signal on

`/v3/wireless-module/<id>/stop`

It then outputs the recorded data and battery stats into a CSV file for the specified module. 

## Screenshots

Screen Shot 2020-01-27 at 1 30 28 pm

Prints when certain modules are started and stopped. When it is stopped it is saved.

## Steps to Test

To run the logger use `python3 mqtt_wireless_log.py`
This will save the logs to the directory where you run the command

### DIFF
--- a/Raspi/Test/DataToTempCSV.py
+++ b/Raspi/Test/DataToTempCSV.py
@@ -23,7 +23,7 @@ class DataToTempCSV:
             self.type = "BATTERY"
             self.parse_module_battery()
 
-        elif msg.topic == "v3/wireless-sensor/battery/low":
+        elif msg.topic == "/v3/wireless-module/battery/low":
             self.type = "BATTERY_LOW"
             self.parse_low_battery()
 
--- a/Raspi/Test/DataToTempCSV.py
+++ b/Raspi/Test/DataToTempCSV.py
@@ -12,35 +12,6 @@ def DataToTempCSV(msg, module_start_time, module_id):
     module_start_time:          Start time of the module (datetime obj)
     """
 
-    data_dict = {}             # Data to be output to a temp CSV
-
-    # Decode the data as utf-8 and load into python dict
-    module_data = msg.payload.decode("utf-8")
-    module_data = json.loads(module_data)
-
-    # Determine which type of data to parse
-    if msg.topic.endswith("data"):
-        type = "DATA"
-        parse_module_data()
-
-    elif msg.topic.endswith("battery"):
-        type = "BATTERY"
-        parse_module_battery()
-
-    elif msg.topic == "/v3/wireless-module/battery/low":
-        type = "BATTERY_LOW"
-        # Nothing to parse
-
-    # Find the difference in seconds to when the recording was started and
-    # when the data was recieved.
-    time_delta = datetime.now() - module_start_time
-    time_delta = time_delta.total_seconds()
-    time_dict_key = f"{module_id}_{type}_TIME"
-    data_dict[time_dict_key] = time_delta
-
-    # Add or create the temp CSV to store the data
-    self.make_temp_csv()
-
     def parse_module_data(self):
         """ Parses the module data if it is from the sensors """
 
@@ -87,3 +58,32 @@ def DataToTempCSV(msg, module_start_time, module_id):
 
             # Append the data onto the temporary file
             csv_writer.writerow(self.data_dict)
+
+    data_dict = {}  # Data to be output to a temp CSV
+
+    # Decode the data as utf-8 and load into python dict
+    module_data = msg.payload.decode("utf-8")
+    module_data = json.loads(module_data)
+
+    # Determine which type of data to parse
+    if msg.topic.endswith("data"):
+        type = "DATA"
+        parse_module_data()
+
+    elif msg.topic.endswith("battery"):
+        type = "BATTERY"
+        parse_module_battery()
+
+    elif msg.topic == "/v3/wireless-module/battery/low":
+        type = "BATTERY_LOW"
+        # Nothing to parse
+
+    # Find the difference in seconds to when the recording was started and
+    # when the data was recieved.
+    time_delta = datetime.now() - module_start_time
+    time_delta = time_delta.total_seconds()
+    time_dict_key = f"{module_id}_{type}_TIME"
+    data_dict[time_dict_key] = time_delta
+
+    # Add or create the temp CSV to store the data
+    self.make_temp_csv(module_data)
--- a/Raspi/Test/DataToTempCSV.py
+++ b/Raspi/Test/DataToTempCSV.py
@@ -35,7 +35,8 @@ class DataToTempCSV:
         # when the data was recieved.
         self.time_delta = datetime.now() - self.module_start_time
         self.time_delta = self.time_delta.total_seconds()
-        self.data_dict[self.module_id+"_"+self.type+"_TIME"] = self.time_delta
+        time_dict_key = f"{self.module_id}_{self.type}_TIME"
+        self.data_dict[time_dict_key] = self.time_delta
 
         # Add or create the temp CSV to store the data
         self.make_temp_csv()
@@ -52,9 +53,8 @@ class DataToTempCSV:
 
             if isinstance(sensor_value, dict):
                 # For nested sensor values
-                for sub_sensor in sensor_value.keys():
+                for (sub_sensor, sub_sensor_value) in sensor_value.items():
                     sub_sensor_name = sensor_name + '_' + sub_sensor
-                    sub_sensor_value = sensor_value[sub_sensor]
                     self.data_dict[sub_sensor_name] = sub_sensor_value
             else:
                 self.data_dict[sensor_name] = sensor_value
--- a/Raspi/Test/DataToTempCSV.py
+++ b/Raspi/Test/DataToTempCSV.py
@@ -2,7 +2,7 @@ import json
 from datetime import datetime
 import os
 import csv
-# mosquitto_pub -t "/v3/wireless-module/battery/low" -m '{"module-id": 2}'
+
 
 class DataToTempCSV:
     """ Single use object to parse the MQTT data and convert it to a temporary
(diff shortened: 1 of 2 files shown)


khanguslee (commented):
Is it me or is the mock wireless sensor scripts gone?

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+
+        elif msg.topic == "/v3/wireless-module/battery/low":
+            self.type = "BATTERY_LOW"
+            # Nothing to parse
+
+        # Find the difference in seconds to when the recording was started and
+        # when the data was recieved.
+        self.time_delta = datetime.now() - self.module_start_time
+        self.time_delta = self.time_delta.total_seconds()
+        self.data_dict[self.module_id+"_"+self.type+"_TIME"] = self.time_delta
```
khanguslee: I would put this into another variable `dict_key = self.module_id+"_"+self.type+"_TIME"`. Don't be afraid to use another line of code if it means the code is easier to read!

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+                    sub_sensor_name = sensor_name + '_' + sub_sensor
+                    sub_sensor_value = sensor_value[sub_sensor]
+                    self.data_dict[sub_sensor_name] = sub_sensor_value
+            else:
+                self.data_dict[sensor_name] = sensor_value
+
+    def parse_module_battery(self):
+        """ Parses the module data if it is from the battery """
+
+        self.data_dict[self.module_id + "_percentage"] = \
```
khanguslee: Would also put this into it's own variable too.

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -0,0 +1,152 @@
...
+
+def on_message(client, userdata, msg):
+    """ MQTT callback for when data is sent on the subscribed
+    '/v3/wireless-module/#' topics. The module_id is found by squashing M
+    infront of the module number. eg M1, M2, M3... etc. The module_id is used
+    for identifying what data came from where and is also used for naming the
+    temp files.
+    """
+
+    module_id = "M" + str(msg.topic[20])
```
khanguslee: I wouldn't hard code the number 20 for this case. Instead, try using the string `split()` function to parse the topic string via `/`. You should get an array of strings that were separated by the `/`.

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Start the recording of <module_id>
+    if msg.topic.endswith("start"):
+        # Clear the old temporary files in the folder, just in case the script
```
khanguslee: Debating whether to put the things inside the if statements (apart from the prints) into their own functions. Thoughts?

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+if __name__ == "__main__":
+    # Global dicts to store state
+    # Dict structure is {<module_id> : <data>}
+    is_recording = {}       # If the data is being recorded
+    module_start_time = {}  # When the data started being recorded
+    output_filename = {}    # Output filename
+
+    broker_address = 'localhost'
```
khanguslee: Definitely would like this to be arg parsed instead.

khanguslee (changes requested):
Some small changes. Can't test this PR because the mock wireless sensor code is missing? Your python code has improved a lot btw! Good use of commenting which made it very easy to read and understand your code!

Raspi/Test/DataToTempCSV.py
emilytrau: This would be a good place to use fstrings https://cito.github.io/blog/f-strings/

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+            self.module_data["percentage"]
+
+    def make_temp_csv(self):
+        """ Makes a temporary CSV file that is hidden and is in the form of
+        .~temp_<filename>.csv in the current directory"""
+
+        # Ensures that the temp file is in the same folder as the script
+        current_dir = os.path.dirname(__file__)
+        temp_filename = str('.~temp_' + self.module_id+'_'+self.type + '.csv')
```
emilytrau: You don't need to do a string conversion here

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+        temp_file_path = os.path.join(current_dir, temp_filename)
+
+        # Find the names for the CSV columns
+        column_names = []
+        for key in self.data_dict.keys():
+            column_names.append(key)
```
emilytrau: This is the same as `column_names = self.data_dict.keys()`

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Merge the dataframes and output the final csv
+    merged_dataframe = pd.concat(temp_dataframes, axis=1)
+    merged_dataframe.to_csv(output_filename)
+
+
+if __name__ == "__main__":
+    # Global dicts to store state
+    # Dict structure is {<module_id> : <data>}
+    is_recording = {}       # If the data is being recorded
```
emilytrau: These variables can be declared at the top of the script for clarity. In general it's a bad idea to define variables below where you use them as it can be hard to find when reading code.

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
+import json
+from datetime import datetime
+import os
+import csv
+
+
+class DataToTempCSV:
```
emilytrau: This could be written as a function instead as you're not keeping the class instance when you call it

emilytrau: It might save you a lot of trouble finding the temp files again if the function returned the temp file name.

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+        # Retrieve sensor data from python dict
+        sensor_data = self.module_data["sensors"]
+
+        for sensor in sensor_data:
+            sensor_name = self.module_id + "_" + sensor["type"]
+            sensor_value = sensor["value"]
+
+            if isinstance(sensor_value, dict):
+                # For nested sensor values
+                for sub_sensor in sensor_value.keys():
```
emilytrau: A neat trick you can do to get the key and value at the same time: `for (sub_sensor, sub_sensor_value) in sensor_value.items()`

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -0,0 +1,152 @@
...
+    # Record low battery data
+    elif msg.topic == "/v3/wireless-module/battery/low":
+        module_data = msg.payload.decode("utf-8")
+        module_data = json.loads(module_data)
+        module_id = "M" + str(module_data["module-id"])
+
+        DataToTempCSV(msg, module_start_time[module_id], module_id)
+
+    # Record other data (battery and sensor data)
+    elif is_recording[module_id] is True:
```
emilytrau: Since `is_recording[module_id]` is a boolean (more specifically it is [truthy](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false)) you can drop the `is True` part.

emilytrau (changes requested):

emilytrau:
> Is it me or is the mock wireless sensor scripts gone?

@Blake-Haydon You can run a `git merge master` to bring this branch up to date with the latest changes (and bring in the test script).

Raspi/Test/DataToTempCSV.py
```diff
@@ -0,0 +1,92 @@
...
+            else:
+                self.data_dict[sensor_name] = sensor_value
+
+    def parse_module_battery(self):
+        """ Parses the module data if it is from the battery """
+
+        self.data_dict[self.module_id + "_percentage"] = \
+            self.module_data["percentage"]
+
+    def make_temp_csv(self):
```
emilytrau: If you wrote the csv data to [a string](https://stackoverflow.com/a/9157370) instead you could probably skip writing a temporary file altogether. The current way works so this is optional

Raspi/Test/DataToTempCSV.py
Blake-Haydon: I didn't realise that python had fstrings. I will definitely use them instead as it's hard to read.

Blake-Haydon:
> Is it me or is the mock wireless sensor scripts gone?

I think the mock wireless sensor scripts are gone because I branched before the commit when they were added. I'll try and get them when I update this pull request

Raspi/Test/DataToTempCSV.py
Blake-Haydon: The temp files are always stored in the directory where the script is run. For example when the script is run in /Das/Test it will be saved in the directory

Raspi/Test/DataToTempCSV.py
Blake-Haydon: The main reason I used temporary csvs is that if the stop message is never received the script wont write the merged csvs with all of the information about the module. If the data is needed can just look at the temp csvs if the data is needed.

Raspi/Test/mqtt_wireless_log.py
Blake-Haydon: It would probably make it nicer to read but for stop it is literally 2 lines of code.

Blake-Haydon: I'll try it and see how it looks. It is a really good point though, I was tempted when I first wrote it.

emilytrau: I think it's better as it is. Putting it in a function is another layer of indirection and it's already pretty clear what the purpose of the code block is.

emilytrau (approved):
Looks good

hallgchris (approved):
Looks good!

khanguslee (approved):
LGTM. Noticed that, the mock wireless module script generates the filename. I think the logger that we have here in `mqtt_wireless_log.py` should control this instead. Log files should not have their filename based on time as well since there won't be an accurate clock on the RPI due to no internet connectivity.

maanika (approved):

PR #43, opened 2020-03-01
DAS.js to collect power using a 3 second rolling average

emilytrau:
## Description

Implement a 3 second rolling average on the power statistic to smooth out spikes.

Not tested yet.

## Screenshots

N/A

## Steps to Test

Start DAS.js, connect the ant+ dongle and pedals, and view the MQTT output. The outputted values should be an average of the raw power values logged.

### DIFF
--- a/Raspi/utils/average.js
+++ b/Raspi/utils/average.js
@@ -30,9 +30,9 @@ class RollingAverage {
   average() {
     // Remove all stale values
     const time = Date.now();
-    this.points = this.points.filter((x) => x.time >= time - this.interval);
+    this.points = this.points.filter((x) => x[0] >= time - this.interval);
 
-    const sum = this.points.reduce((a, b) => a + b, 0);
+    const sum = this.points.reduce((acc, curr) => acc + curr[1], 0);
     const average = sum / this.points.length;
     return average;
   }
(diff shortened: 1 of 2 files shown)


emilytrau:
Need to test this when I get my hands on the pedals.

Raspi/utils/average.js
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Calculate the current average value over the rolling interval
+   *
+   * @returns {number} Average value
+   */
+  average() {
+    // Remove all stale values
+    const time = Date.now();
+    this.points = this.points.filter((x) => x[0] >= time - this.interval);
```
khanguslee, line 33: What are your thoughts on making a fixed length Queue that pushes old data when a new data comes in if the Queue has hit that length. Will be less intensive then filtering an array each time.

khanguslee (changes requested):
Thoughts on using a Queue data structure instead of filtering?

Raspi/utils/average.js
emilytrau: I think it's a non-issue. We're getting data from the pedals at ~1Hz, and we're sending to the broker at 1Hz. At most the buffer will have 5-10 items. I doubt this will become a bottleneck, but we can fix it if it comes to that.

khanguslee: Yup, totally agree. Given that the maximum amount of items in `this.points` will be around ~10 items. Just wanted to raise this point for next time!

khanguslee (approved):
LGTM

PR #46, opened 2020-03-10
Implement sorting into minutes into das data sorting script 

nix-chu:
## Description

-Adapted from nix.chu/sorts-data branch
-Implements a sorter into minutes to accommodate large data 
-Or a sorter into seconds for more precise data

It should be noted that **the branch nix.chu/sorts-data needs to be merged first** before merging this branch, as changes from this branch started from that branch. 

## Steps to Test

Using computer terminal, run `python das_data_sorter.py --file (data file) --output (output file) --unit (seconds/minutes)`
- file requires input
- output default is "filtered_data.csv", unless specified by the user
- unit default is in seconds, unless specified by the user

### DIFF
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -6,8 +6,8 @@ from math import ceil
 parser = argparse.ArgumentParser()
 parser.add_argument("--file", help="Input CSV file", action="store", required=True)
 parser.add_argument("--output", help="Returns the filtered data", default="filtered_data.csv", action="store")
-parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)",
-                     default="seconds", choices=["seconds", "s", "minutes", "m"], action="store")
+parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", 
+                    choices=["seconds", "s", "minutes", "m"], action="store")
 args = parser.parse_args()
 
 # initialising data
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -6,11 +6,12 @@ from math import ceil
 parser = argparse.ArgumentParser()
 parser.add_argument("--file", help="Input CSV file", action="store")
 parser.add_argument("--output", help="Returns the filtered data", default="filtered_data.csv", action="store")
+parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", action="store")
 args = parser.parse_args()
 
 # initialising data
 data = pd.read_csv(args.file)
-milliseconds = data["time"]/1000
+milliseconds = data["time"]
 gps_course = data["gps_course"]
 gps_speed = data["gps_speed"]
 tempC = data["thermoC"]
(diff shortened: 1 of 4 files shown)


utils/das_data_sorter.py
```diff
@@ -0,0 +1,75 @@
+import pandas as pd 
+import argparse 
+from math import ceil
+
+# accepts terminal commands
+parser = argparse.ArgumentParser()
+parser.add_argument("--file", help="Input CSV file", action="store")
+parser.add_argument("--output", help="Returns the filtered data", default="filtered_data.csv", action="store")
+parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", action="store")
```
khanguslee: If you add `choices` you can limit the values you input into an argument.

`parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", choices=['seconds', 's', 'minutes', 'm'], action="store")`

utils/das_data_sorter.py
```diff
@@ -0,0 +1,75 @@
...
+tempC = data["thermoC"]
+tempF = data["thermoF"]
+cadence = data["cadence"]
+power = data["power"]
+reed_vel = data["reed_velocity"]
+reed_dis = data["reed_distance"]
+
+time = []
+if args.unit == "seconds" or args.unit == "s":
+    time = milliseconds/1000
```
khanguslee, line 41: It doesn't make sense to overwrite the time variable with an integer if you declare it as an empty array above. Maybe you should rename the variable and declare it as `None`?

khanguslee (changes requested):
Need to wait for the other PR to get pulled in first.

utils/das_data_sorter.py
nix-chu: Ah, the milliseconds variable is an array of data, not an integer. So at the end of this code block, time variable becomes an array of data (divided by 1000) as well.

khanguslee:
Please update this branch with `master` branch!

nix-chu:
@khanguslee updated with master branch!

utils/das_data_sorter.py
khanguslee: ohhhh, yup. I misread that thank you. data is in a dataframe.

utils/das_data_sorter.py
```diff
@@ -20,39 +23,48 @@
...
-seconds.append(millisecond_indices)
+
+new_time.append(millisecond_indices)
 
 # averages data within the seconds, according to seconds array 
 def avg_Data(array):
     avg_array = []
-    for second in seconds:
+
+    for t in new_time:
```
khanguslee: Had to debug through this code to understand what was going on. I think I just didn't know that `new_time` was a list of lists. Can you think of renaming the `t` variable to something more meaningful? Suggestion: `current_time_indices`? Not too sure tbh. But something like that would be helpful!

utils/das_data_sorter.py
```diff
@@ -20,39 +23,48 @@
...
         total = 0
-        for i in second:
+        for i in t:
```
khanguslee: I would also rename `i` to something else. `actual_millisecond`? Your decision! Hope you get the idea.

khanguslee (changes requested):
Some changes to the way you name some of the variables. Something like `i` and `t` are okay when you are using a for loop and the variable represents an index number. Since your `i` and `t` variables actually hold something more 'meaningful', I would rename them. If you had to describe that variable in 3 words or less, what would you say?

I would like to run unit tests and remove global variables in your script later on! But for now, I'd like to make small steps! Good work on your second PR!

Screenshot from my own testing: Left is seconds, right is minutes. Can confirm the script works.
Screen Shot 2020-04-19 at 10 52 18 pm

utils/das_data_sorter.py
```diff
@@ -2,15 +2,31 @@
 import argparse 
 from math import ceil
 
+# averages data within the seconds, according to seconds array 
+def avg_Data(array):
```
khanguslee, line 6: Curious as to why you moved the function to the top? Generally, this is a good thing. But since you are using a global variable `new_time`, I'm not sure how Python will react if you were to run the `avg_Data` function before `new_time` is defined. 

This problem would be solved if we didn't have such global variables, but we'll slowly teach you some new tricks ;)

nix-chu: I just ran it, and it worked just as fine. But I understand what you meant by that, it would be a problem in some cases. I put the function on top because I remembered it being good practice during my Javascript days.

khanguslee (approved):
LGTM

PR #48, opened 2020-04-14
Fixed output csvs to display a unique ID rather than the time

Blake-Haydon:
## Description

+ Moved the output files to the log folder
+ Files are now in the format "{ID}_{Module ID}.csv"
+ Cleaned up the pipenv files to include the new dependencies

## Screenshots

image
Shows the new file position and format

## Steps to Test

1. Run Raspi/Test/mqtt_wireless_log.py
2. After that has connected to MQTT brocker run Raspi/Test/mock_wireless_module.py -t 5

Hopefully the logger should record the data in the log folder!


Raspi/Test/Pipfile
```diff
@@ -1,13 +1,16 @@
...
+verify_ssl = true
+
+[dev-packages]
 
 [packages]
 paho-mqtt = "*"
-"urllib3" = "*"
-
-[dev-packages]
+scipy = "*"
```
hallgchris, line 10: What do we use `scipy` for?

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -154,7 +168,7 @@ def merge_temps(output_filename, temp_filepaths):
 
     # Merge the dataframes and output the final csv
     merged_dataframe = pd.concat(temp_dataframes, axis=1)
-    merged_dataframe.to_csv(output_filename)
+    merged_dataframe.to_csv("../logs/" + output_filename)
```
hallgchris: This only works if `mqtt_wireless_log.py` is run from the `Test` folder, on my machine at least. Otherwise, `to_csv` fails, unable to find the `logs` folder. I think this would be fixed by using something relative to `os.path.dirname(__file__)` like in `DataToTempCSV.py`. (also, this failure seems to be silent, we should probably alert the user if something like this happens.)

This also made me realise, should `mqtt_wireless_log.py` really be in the `Test` folder? It's not for testing, it's our actual code to be run on the DAS. It also feels a bit strange for the code in one subfolder to be creating files in another, seemingly unrelated subfolder. Maybe this should be a separate PR though.

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -70,17 +72,29 @@ def on_message(client, userdata, msg):
...
 
     # Save the state of recording and the output filename to global dicts
     is_recording[module_id_str] = True
     module_start_time[module_id_str] = datetime.now()
 
-    # Generate filename from the current date and time in the Australian format
-    output_filename[module_id_str] = module_start_time[module_id_str].strftime(
-        module_id_str + "_D%d-%m-%d_T%H-%M-%S" + ".csv")
+    # Generate filename from the last log number + 1
+    max_file_id = 0
```
hallgchris, line 84: This section is probably a great candidate for unit testing. Code that could be easy to screw up in the future, and also has scary consequences (loss of data).

You could separate this into a function that accepts a list of file names, and returns a correct new filename. Then it can be called here and tested from elsewhere

hallgchris (changes requested):

hallgchris:
Oh also, congrats on first PR this side of the break!

Raspi/Test/mqtt_wireless_log.py
khanguslee: Definitely agree that this file shouldn't be in Test. It's actually a core component of the DAS. So should really be in the root directory. Definitely need to plan out how we want to setup this repo for maintanability purposes.

Raspi/Test/mqtt_wireless_log.py
khanguslee: Next PR, should be a unit test solely to protect edge cases are not happening.

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -70,17 +72,29 @@ def on_message(client, userdata, msg):
...
+    for filepath in glob.glob('./../logs/*_M?.csv'):
```
khanguslee: What does the solution look like if we were to instead just iterate through the files that are INSIDE this directory? We wouldn't need to split the filename below.

Can you also put the directory path / regex as a variable?

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -70,17 +72,29 @@ def on_message(client, userdata, msg):
...
+        # split the filepath into the filename
+        filename = filepath.split("/")[3]
+
+        # Gets all the digits from the file name
+        temp = re.findall(r'\d+', filename)
+
+        # Gets all the digits from the file name
+        file_id = list(map(int, temp))[0]
```
khanguslee, line 93: As @hallgchris mentioned above, I think i'd be comfortable if this part of the code was unit tested somehow. (Definitely in your next PR since it's quite new to you)

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -70,17 +72,29 @@ def on_message(client, userdata, msg):
...
+
+        if file_id > max_file_id:
+            max_file_id = file_id
+
+    output_filename[module_id_str] = f"{max_file_id+1}_{module_id_str}.csv"
```
khanguslee, line 98: This will fail on the RPI as f strings are a version 3.6 thing. I'm willing to let this go, as i would try to install version 3.6+ on the RPI.

Raspi/Test/Pipfile
```diff
@@ -1,13 +1,16 @@
...
+pandas = "*"
+argparse = "*"
+urllib3 = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.8"
```
khanguslee, line 16: Is there a reason why this was upgraded to 3.8? We might need to look into setting up `pyenv` on everyones machines.

khanguslee (changes requested):
Just a few concerns, but the PR looks good in general! Mate, your python code has really improved since you started ;)

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -70,17 +72,29 @@ def on_message(client, userdata, msg):
...
     module_start_time[module_id_str] = datetime.now()
 
-    # Generate filename from the current date and time in the Australian format
-    output_filename[module_id_str] = module_start_time[module_id_str].strftime(
-        module_id_str + "_D%d-%m-%d_T%H-%M-%S" + ".csv")
+    # Generate filename from the last log number + 1
+    max_file_id = 0
+    for filepath in glob.glob('./../logs/*_M?.csv'):
+        # split the filepath into the filename
+        filename = filepath.split("/")[3]
```
emilytrau: This assumes that the filename is always the 4th element. `[-1]` would be a safer way to get the last element.

Raspi/Test/mqtt_wireless_log.py
emilytrau: It might be a good idea to clean up the temp files after you're done with them.

emilytrau (changes requested):

Raspi/Test/Pipfile
Blake-Haydon: Oops I think I must have installed it by accident

Raspi/Test/mqtt_wireless_log.py
Blake-Haydon: Yeah this is something that I will look into after finish this PR, its kind of a mess in there

Blake-Haydon: Yeah they get cleaned up in the main body of the code

Blake-Haydon:
It seems as if I have fixed the issues with files not saving.

I will write unit tests once the DAS has been moved into a new file layout.

khanguslee:
I ran `python mqtt_wireless_log.py` on left terminal and `python mock_wireless_module.py -t 5` on the right terminal. Doesn't seem like the logger is merging the files? 

I also have a bunch of temp files in my folder now. 

Screen Shot 2020-04-22 at 6 41 26 pm

Blake-Haydon:
This is very weird, let me run it again.

Raspi/Test/mqtt_wireless_log.py
```diff
@@ -154,7 +169,8 @@ def merge_temps(output_filename, temp_filepaths):
 
     # Merge the dataframes and output the final csv
     merged_dataframe = pd.concat(temp_dataframes, axis=1)
-    merged_dataframe.to_csv(output_filename)
+    file_path = os.path.dirname(__file__)
```
hallgchris, line 172: I haven't looked into the cause, but it seems that when you run `python mqtt_wireless_log.py` from inside the `Test` directory, it fails to save the files because `os.path.dirname(__file__)` is for some reason returning an empty string, causing python to panic and stop everything.

It works fine if you run the script with `python Test/mqtt_wireless_log.py` from the `Raspi` directory.

hallgchris (changes requested):

Blake-Haydon:
Just fixed up the merging function, it should work now

PR #50, opened 2020-04-20
Rewrite DAS Data Sort script into Object Orientated Class

nix-chu:
## Description
- Purpose of script is to convert a group of data points into one singular data point, based on the time unit provided by user. 

1. Reads the given CSV file and the data columns.
2. Combines the milliseconds within the given time intervals (seconds/minutes) into a singular time unit.
3. Then, the corresponding data points are averaged within each time interval. Excludes gps data, which will return a zero if the gps was turned off at this time, or a one if the gps was turned on. 
4. Creates a new CSV file and prints the newly manipulated data.

## Steps to Test
Only CSV files made after Ford 2019 will work with this script. I am currently using data_291.csv to test my script, which can be found on this thread. 
1. In the utils folder, transfer one of the CSV data files from the DAS into this folder. 
2. In the terminal, run `python das_data_sorter --input (file name) --output (output name) --unit (seconds/minutes)`
3. When finished, it should print 'Success!' into the terminal.


nix-chu:
For lines 12 and 27, let me know if rounding on another line is unconventional. I did this because I thought line 11 and 26 looked messy enough.

Also, let me know if rounding to two decimal places is also sufficient.

khanguslee:
BTW, you can reference other PRs by putting the full link of the PR here OR (if it's the same repo, use #<number>) see https://help.github.com/en/github/writing-on-github/autolinked-references-and-urls for an example.

khanguslee:
Next time, you can add a comment to a line in your code. So it's easier for the reviewer to look at these lines of code!

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
 import pandas as pd 
 import argparse 
 from math import ceil
+from numpy import median, mean
+
+# averages data within the seconds or minutes, according to new_time array
+def avg_Data(array):
```
khanguslee: Was it intentional to have two functions called `avg_Data` ? It also looks like theres a small difference in both of them too, so which one did you want?

Also forgot to say this last time, can you rename this method to `average_data`. See https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables on how to name methods using the PEP 8 python conventions.

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
...
+def mean_smooth(data):
+    new_data = []
+
+    for i in range(1,len(data)-1):
+        first = data[i-1]
+        second = data[i]
+        third = data[i+1]
+
+        new_data_point = mean([first, second, third])
+        new_data_point = round(new_data_point, ndigits=2)
```
khanguslee: I'd like to control the number of digits to round. Can you add a new `precision=2` argument to the `mean_smooth` function (By adding `=2`, we set a default parameter if the user does not include a second parameter to this function.

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
...
+        new_data_point = median([first, second, third])
+        new_data.append(new_data_point)
+    
+    return new_data
+
+# smoothing technique function caller
+def smooth_data(data, technique):
+    if technique == "mean":
+        # when input is "--smooth mean"
+        final_data = {
```
khanguslee: I think I would have preferred if you had iterated through the keys and values of the `data` dictionary and then added it to the `final_data` dictionary. This is because the keys for both dictionaries are the same.

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
...
+        third = data[i+1]
+
+        new_data_point = median([first, second, third])
+        new_data.append(new_data_point)
+    
+    return new_data
+
+# smoothing technique function caller
+def smooth_data(data, technique):
+    if technique == "mean":
```
khanguslee: The if statements are quite unnecessarily long. When I look at the two if statements, I can see that they are very similar. The only difference is that we use a different calculation method depending on the technique. Try assigning the technique function to a variable so it is cleaner.

Here is an example of how I would resolve this problem.

``` python
smoothing_method = None
if technique is "mean":
	smoothing_method = mean_smooth
elif technique is "median":
	smoothing_method = median_smooth
else:
	raise ValueError("Invalid technique given")

final_data = {
            "time": range(1, len(new_time)-1),
            "gps_course": smoothing_method(data["gps_course"]),
            "gps_speed": smoothing_method(data["gps_speed"]),
            "tempC": smoothing_method(data["tempC"]),
            "tempF": smoothing_method(data["tempF"]),
            "cadence": smoothing_method(data["cadence"]),
            "power": smoothing_method(data["power"]),
            "reed_velocity": smoothing_method(data["reed_velocity"]),
            "reed_distance": smoothing_method(data["reed_distance"])
        }
return final_data
```

utils/das_data_sorter.py
```diff
@@ -21,6 +94,9 @@ def avg_Data(array):
 parser.add_argument("--output", help="Returns the filtered data", default="filtered_data.csv", action="store")
 parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", 
                     choices=["seconds", "s", "minutes", "m"], action="store")
+
+parser.add_argument("--smooth", help="Smoothens data points by 3-point mean or median smoothing",
```
khanguslee: What happens if a user does not input a `--smooth` argument?

utils/das_data_sorter.py
```diff
@@ -72,5 +148,15 @@ def avg_Data(array):
     "power": avg_Data(power),
     "reed_velocity": avg_Data(reed_vel),
     "reed_distance": avg_Data(reed_dis)
-})
+}
+
+# checks if user wants to smooth data using moving mean or moving median method
+# if no input from user, script will ignore this whole block
+if args.smooth == "mean":
```
khanguslee: Seems like a unnecessary if statement here? We can simply just pass in `args.smooth`

khanguslee (changes requested):
Some changes to make, but very minor! This will come in time as you continue programming more and more!

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
...
+    avg_array = []
+
+    for millisecond_indices in new_time:
+        avg = mean(array[millisecond_indices])
+        avg = round(avg, ndigits=2)
+        avg_array.append(avg)
+
+    return avg_array
+
+# 3 point moving mean smoothing
```
hallgchris: When you have a comment which is documenting a function, we usually use python's multiline comments, i.e.
```python
def mean_smooth(data):
""" 3 point moving mean smoothing """
    new_data = [] # etc.
```
If you do your function documentation like this, whatever editor will recognise it and display it as the description of a function when you mouse over it (for example)
image

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
...
+def median_smooth(data):
+    new_data = []
+
+    # iterates through each data point and its adjacents (excluding first and last) and finds the median 
+    for i in range(1, len(data)-1):
+        first = data[i-1]
+        second = data[i]
+        third = data[i+1]
+
+        new_data_point = median([first, second, third])
```
hallgchris: Do we not want to round when we're calculating medians?

utils/das_data_sorter.py
hallgchris: In general when coding, if you've ever had to type the same thing twice, there's almost always a better way :wink:

utils/das_data_sorter.py
```diff
@@ -1,6 +1,79 @@
 import pandas as pd 
 import argparse 
 from math import ceil
+from numpy import median, mean
+
+# averages data within the seconds or minutes, according to new_time array
+def avg_Data(array):
+    avg_array = []
+
+    for millisecond_indices in new_time:
```
hallgchris: Here we're accessing the variable `new_time`, which isn't defined until much later in the file. Python allows this, but in most cases I'd try to avoid accessing variables outside of the scope of the function.

The reason for this is mainly because it makes code easier to read, and easier to debug. When I first read this code, I had no clue what this variable was, but I would have a better idea if it was instead a parameter of the function. As for debugging, functions are much easier to debug when they are deterministic - meaning that for the same inputs, they will always return the same outputs. This is not the case here, as the function's output also depends on the value of `new_time` which is completely independent.

`new_time` appears in a few other functions around this file, they could be made parameters too!

hallgchris (changes requested):
Just a few things to add on to Angus's feedback! On the whole looking really good though

utils/das_data_sorter.py
nix-chu: How do I specify the names for the keys (which represent each column of data without hardcoding it?

utils/das_data_sorter.py
nix-chu: I suppose in this case, the user would just want normal converted data. Some users might want this, so that they can know the exact values at certain points of time.

utils/Pipfile
```diff
@@ -9,6 +9,7 @@ verify_ssl = true
 pandas = "*"
 argparse = "*"
 math = "*"
```
hallgchris: What is the `math` package, is it meant to be the one built into python? It's causing issues for me trying to `pipenv install`. I think this line needs to be removed.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
-    return avg_array
-
-# accepts terminal commands
-parser = argparse.ArgumentParser()
+# accepts terminal arguments
+parser = ArgumentParser()
 parser.add_argument("--file", help="Input CSV file", action="store", required=True)
 parser.add_argument("--output", help="Returns the filtered data", default="filtered_data.csv", action="store")
 parser.add_argument("--unit", help="Specify time units (seconds, s, or minutes, m)", default="seconds", 
                     choices=["seconds", "s", "minutes", "m"], action="store")
```
hallgchris: `output` and `unit` are probably required fields, like `file`.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
-
-        avg = round(total/len(index_array), ndigits=2)
-        avg_array.append(avg)
-    return avg_array
-
-# accepts terminal commands
-parser = argparse.ArgumentParser()
+# accepts terminal arguments
+parser = ArgumentParser()
 parser.add_argument("--file", help="Input CSV file", action="store", required=True)
```
hallgchris: Should call this argument `input`, just for consistency?

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+
+        self.print_output_file()
+
+    def convert_time(self, milliseconds:D) -> D:
+        '''Returns the conversion of the time data points from milliseconds to the specified time unit, depending on the 
+        argument passed by args.unit.
+        
+        args.unit accepts only seconds/s and minutes/m as arguments. 
+        '''
+        if args.unit == "seconds" or args.unit == "s":
```
hallgchris: Ideally, we want to be able to use this `DasSort` class without needing to worry about what variables are set outside of it. Here, we're referencing a variable `args` which might not necessarily exist if we for example created an instance of this class from another python file. (Another example - someone might change `args` without updating `DasSort`, as there's no clear connection between the two unless you read all the code)

Instead, I would pass the different bits of `args` that you need in the constructor of `DasData` and store them to be used whenever, such as here. That way, the variables have a clear obvious association with the class and will always be accessible, regardless of what happens outside the class.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
-    "tempC": avg_Data(tempC),
-    "tempF": avg_Data(tempF),
-    "cadence": avg_Data(cadence),
-    "power": avg_Data(power),
-    "reed_velocity": avg_Data(reed_vel),
-    "reed_distance": avg_Data(reed_dis)
-})
-final.to_csv(args.output, index=False)
+class DasSort:
+    def __init__(self, data:str) -> None:
```
hallgchris: Awesome to see type hinting! The type of `data` should be `pd.DataFrame`, I believe.

image

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+        self.print_output_file()
+
+    def convert_time(self, milliseconds:D) -> D:
+        '''Returns the conversion of the time data points from milliseconds to the specified time unit, depending on the 
+        argument passed by args.unit.
+        
+        args.unit accepts only seconds/s and minutes/m as arguments. 
+        '''
+        if args.unit == "seconds" or args.unit == "s":
+            milliseconds = milliseconds/1000
```
hallgchris: I think we could have a better name for this variable, as the data is no longer in units of milliseconds.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+        Disregards None and zero in calculation. Otherwise, if input is invalid, then raises error.
+        '''
+        total = 0
+        length = len(data_array)
+        for number in data_array:
+            if number == 0 or pd.isna(number):
+                # None and zeroes are ignored
+                length -= 1
+                continue
+            elif number > 0 or number < 0:
```
hallgchris: Unless you've thought of some special case I haven't thought of, if `number == 0` was false, then `number > 0 or number < 0` will always be true. So, this case in the if statement will always run, and the `else` section below will never be reached. You should be able to replace the whole thing with `total += number`

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+                previous_time = ceil(time[i])
+                index_array = [] # recreates new array if previous time has changed
+            index_array.append(i) 
+
+        # pushes the final array when for loop has completed
+        universal_array.append(index_array)
+
+        return universal_array
+
+    def mean(self, data_array:list) -> int:
```
hallgchris: Input type is pd.Series, return type will be `float`.

Although, `list`s will still work in this function. If you wanted to be really fancy you could do `Union(pd.Series, list)` from the typing module, but that isn't necessary as we don't pass this function `list`s from anywhere.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+
+    def location_data(self, data:D) -> D:
+        '''Handles the location data and returns the data point at the first element of each time interval. 
+
+        gps_location requires a separate function, because the values are strings, whereas the other data columns are 
+        integers/floats. Handles it so that it will always return the location data at the first element of each time interval.
+        '''
+        new_location_data = []
+        for index_array in self.indexes:
+            first_element = index_array[0]
```
hallgchris: `first_element_index` would be more descriptive

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+            "thermoC": self.thermoc,
+            "tempF": self.thermof,
+            "pot": self.pot,
+            "cadence": self.cadence,
+            "power": self.power,
+            "reed_velocity": self.reed_velocity,
+            "reed_distance": self.reed_distance
+        })
+        final.to_csv(args.output, index=False)
+        print("Success! Check in the same folder for the new CSV file.")
```
hallgchris: The output file isn't located in the same folder as the input, it's located wherever the user specified. Maybe instead print `Success! Output written to {args.output}`

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+        '''
+        new_location_data = []
+        for index_array in self.indexes:
+            first_element = index_array[0] 
+            new_location = data[first_element]
+            new_location_data.append(new_location)
+        
+        return new_location_data
+    
+    def print_output_file(self) -> None:
```
hallgchris: We "print" something when it appears on the console, but when we're putting stuff into a file, "write" might be more clear

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
-    "power": avg_Data(power),
-    "reed_velocity": avg_Data(reed_vel),
-    "reed_distance": avg_Data(reed_dis)
-})
-final.to_csv(args.output, index=False)
+class DasSort:
+    def __init__(self, data:str) -> None:
+        self.indexes = None
+        self.time = self.convert_time(data["time"])
+        self.gps = self.average_data(data["gps"])
```
hallgchris: As far as I know, `gps` contains either a 1 when gps data is available, and 0 otherwise. Rather than averaging this data, I think it would make most sense to have a 1 if there's any 1s in the original data, and only a 0 if all values are 0s.

hallgchris (changes requested):
Some great work here! Don't forget to mark the PR ready for review if you've made it a draft or tagged it as WIP. And sorry for taking a few days :)

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
-import argparse 
-from math import ceil
+from argparse import ArgumentParser
+from numpy import ceil
+from typing import TypeVar
 
-# averages data within the seconds, according to seconds array 
-def avg_Data(array):
-    avg_array = []
+D = TypeVar('DAS_data') # data type to represent columns of the data in a given file
```
khanguslee: I really like the concept of this! Can we change the name `D` into a more descriptive type?

utils/das_data_sorter.py
khanguslee: yes pls

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+class DasSort:
+    def __init__(self, data:str) -> None:
+        self.indexes = None
+        self.time = self.convert_time(data["time"])
+        self.gps = self.average_data(data["gps"])
+        self.gps_location = self.location_data(data["gps_location"])
+        self.gps_course = self.average_data(data["gps_course"])
+        self.gps_speed = self.average_data(data["gps_speed"])
+        self.gps_satellites = self.average_data(data["gps_satellites"])
+        self.ax = self.average_data(data["aX"])
```
khanguslee: Once we eventually add `pylint`, it will complain about this class because there's too many variables. This is fine for now but we might want to make some helper classes to store these variables instead.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+    def average_data(self, data:D) -> D:
+        '''Returns a column of new data points that has been averaged, based on specified time unit. 
+
+        Finds the average of all data points within the same time interval. Appends this new value into an array, then continues
+        to the next time interval. Once all time intervals have been accounted for, returns the set of new data points.
+        '''
+        avg_data = []
+
+        for index_array in self.indexes:
+            avg = self.mean(data[index_array])
```
khanguslee: I like to name this variable to something like `current_average`. Thoughts?

utils/das_data_sorter.py
```diff
@@ -1,76 +1,170 @@
...
+                # 'number' value is invalid
+                raise ValueError("Data point invalid and is neither zero nor None. The data point was "+str(number)+". ("+str(type(number))+")")
+        
+        try:
+            return total/length
+        except ZeroDivisionError:
+            # in the event where length = 0, due to all the elements in data_array being ignored
+            return 0 
+
+    def average_data(self, data:D) -> D:
```
khanguslee: I just realised it looks like theres a smiley face and a shocked face `:D) -> D:`

utils/das_data_sorter.py
khanguslee: I was about to say this!

utils/das_data_sorter.py
khanguslee: As a reader of this code, I thought this function would print the whole output file. But once I looked at the function contents, it does something else. Try describe this function in a few words about what it does. That should help you in naming this function!

khanguslee (changes requested):
The code looks so much better than what you had before! Nearly there!

utils/das_data_sorter.py
hallgchris: The function takes a happy face and makes it sad :(

utils/Pipfile
nix-chu: @hallgchris Oops, forgot about the pipfile. I didn't change this file and remove the math package because I didn't know what to do with it, since we are moving on to poetry. Should I delete this file instead?

utils/das_data_sorter.py
nix-chu: I had a problem earlier where, in the CSV file, if the cell was empty, it would say that I was trying to perform addition with an integer and a string. (Shown below is an outdated screenshot)
image

So I thought that an empty cell in Python would read as "". However, after some debugging, I found out that an empty cell's data type is `float`, so I'm just all-round confused. Hence why `pd.isna(number)` is there, which is Pandas's way of handling empty cells. I also kept the else code block there for future debugging. 

(Shown below is the error message when elif statement has `number == ""`)
image

The link below is a reference I found on how to handle empty cells in Python, scroll all the way down to the "Dealing with nan without getting a headache" header:
https://towardsdatascience.com/navigating-the-hell-of-nans-in-python-71b12558895b

utils/das_data_sorter.py
nix-chu: Not sure how to do that, I haven't heard of a helper class before

utils/das_data_sorter.py
nix-chu: I didn't make these required because I have put a default argument for these. 
Output default = `filtered_data.csv`
Unit default = `seconds`
Is this fine or still change it?

utils/das_data_sorter.py
nix-chu: I understand what you are saying, and it would make the script run quicker if it didn't use the average_data method. One problem is that I'm not sure how to go about this. 

When I tried checking if there was a 1 within the time interval, by doing `if 1 in data[index_array]`, it didn't work the way it's supposed to work. After doing some testing on a different Python file, I found that the data points inside the gps column is **not** of the integer type
image

utils/das_data_sorter.py
nix-chu: I got rid of this hahaha. I used this to represent columns of data points, but turns out they have their own data type as discovered by Chris. 

Just copied this concept from my FIT1008 classes, but still unsure how to fully use this. Might ask my lecturer before utilising this.

utils/das_data_sorter.py
khanguslee: @hallgchris why is it `average_data`? Why can't it be `happy_data`?

khanguslee (dismissed):
LGTM. Keen to see what @hallgchris says!

utils/das_data_sorter.py
nix-chu: So according to my lecturer, using TypeVar is like using pronumerals in maths. For a math function like; f: R->R, f(x)=x^2+x+3, this function only allows real numbers to substitute for x. But in Python, TypeVar makes it so that x must be of the same data type throughout the script. This is to allow flexibility in what data types the user inputs.

If the Python function is `def function(arg1: T, arg2 :T)` then that means the argument passed by both arg1 and arg2 must both be strings. Or in another case, both must be integers. Depends on how the user wants to use the class. This ensures that the user is using the same data types for the class. 

For example, we cant do `function("str", 48)`, because T is confused whether it is a string or an integer. But we can do `function("str", "str")` or `function(48, 32)`. 

Upon saying that, she suggests that you only use TypeVar if the type to be used for all scripts is unknown. Otherwise, there's no point in using TypeVar if the arguments passed are intended for a specific data type, because you can just do `def function(arg1: int, arg2: int)`.

utils/das_data_sorter.py
hallgchris: That's odd, it seems to be working for me
image

utils/das_data_sorter.py
hallgchris: Oh yeah that's fine then :)

utils/das_data_sorter.py
```diff
@@ -162,9 +171,9 @@ def print_output_file(self) -> None:
             "reed_velocity": self.reed_velocity,
             "reed_distance": self.reed_distance
         })
-        final.to_csv(args.output, index=False)
-        print("Success! Check in the same folder for the new CSV file.")
+        final.to_csv(file_output, index=False)
+        print("Success! Output is written to {file_output}")
```
hallgchris: You'll need to use backticks (``` ` ` ```) rather than quotes to use python's f-strings. (have a google if you're unfamiliar, they're super useful)

utils/das_data_sorter.py
hallgchris: Actually - I don't think we should have a default output, as it could lead to accidentally overwriting a file. We should either have no default, or have some confirmation about whether you really want to overwrite. (The former would be much easier)

hallgchris (changes requested):
I've just realised an issue - Sometime since ford last year, we replaced the `gps_location` column with three columns, `gps_lat`, `gps_long` and `gps_alt`. As a result, this script crashes when trying to process a log made since around the middle of last year. I think what we need is for this script to see if the `gps_location` column exists, and if it does, process that column, otherwise process the newer three columns. (should also document why you do this, as it may not be clear looking back on the code).

GitHub doesn't seem to let me upload a `.csv`, so I'll send you a log via slack DM

hallgchris:
Nevermind, when I tried to upload it outside of the review environment I got a better error message, just needed to be zipped to make it happy. This is from Al's run from Holden this year :)

[data_291.zip](https://github.com/monash-human-power/data-acquisition-system/files/4639509/data_291.zip)

utils/das_data_sorter.py
nix-chu: Can't seem to use backticks, but doing `print(f"Success {file_output}")` seems to work well ¯\\\_(ツ)\_/¯ ahahaha

hallgchris: Sorry lol I'm confusing python with JS :sweat_smile: That's great now

utils/das_data_sorter.py
```diff
@@ -1,76 +1,179 @@
...
+            "cadence": self.cadence,
+            "power": self.power,
+            "reed_velocity": self.reed_velocity,
+            "reed_distance": self.reed_distance
+        })
+        final.to_csv(file_output, index=False)
+        print("Success! Output is written to {file_output}")
+
+if __name__ == '__main__':
+    data = pd.read_csv(args.input)
```
khanguslee: This looks so much neater than the old script! I can now read at a high-level as to what and how this program works.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,179 @@
...
+        self.gz = self.average_data(file_input["gZ"])
+        self.thermoc = self.average_data(file_input["thermoC"])
+        self.thermof = self.average_data(file_input["thermoF"])
+        self.pot = self.average_data(file_input["pot"])
+        self.cadence = self.average_data(file_input["cadence"])
+        self.power = self.average_data(file_input["power"])
+        self.reed_velocity = self.average_data(file_input["reed_velocity"])
+        self.reed_distance = self.average_data(file_input["reed_distance"])
+
+        self.write_to_output_file(file_output)
```
khanguslee: This should not be ran inside the class instructor. This should be ran by the class instance eg. `das_sort.write_to_output_file(args.output)`. If you do this, you also don't need to include the `file_output` argument.

utils/das_data_sorter.py
```diff
@@ -1,76 +1,179 @@
...
+            "power": self.power,
+            "reed_velocity": self.reed_velocity,
+            "reed_distance": self.reed_distance
+        })
+        final.to_csv(file_output, index=False)
+        print("Success! Output is written to {file_output}")
+
+if __name__ == '__main__':
+    data = pd.read_csv(args.input)
+    das_sort = DasSort(data, args.output, args.unit)
```
khanguslee: Once you add the above changes. This will basically read as:
1. Load csv file
2. Sort csv file
3. Write csv file to a file

If a developer wants to know _how_ the data is sorted, they will read `DasSort` implementation. I hope this gives you an idea about why we would do this!

khanguslee (changes requested):
Some small changes that I also did not pick up last time oops. It's a really small change!

khanguslee:
Also, your commit messages are really good btw! You really got the hang of it! Just get into the habit of committing with one feature but I don't always follow that too ;)

nix-chu:
Some very vital changes made in the last few commits, so I will summarise them so that each key change can be reviewed. I also think that it'd be a good idea to discuss these and gather opinions about the code changes that I made:
1. Was able to fix gps_data method. 

In pandas, for some reason `if 1 in data[index_array]` (line 132) takes the indexes. However `if 1 in data[index_array].values` works, because it was accessing the actual array of values. So this returns whether the GPS was off (0) or on (1) at each time interval, which is what it's intended to do 👍 Perfect!

2. Moved the write_to_output_file outside the constructor, previously it was the last line in `__init__`. 

Angus' logic makes sense, this allows the user to take each step in the process one-by-one. Firstly, read the CSV file, then filter the data, and finally write the filtered data onto a new file. Aside from this, we can adapt the code in the future to allow the user to manipulate the data before writing it onto a new file.

My initial idea for this script was to automate everything for the user, but this is better because it allows flexibility in how the user wants the data to be _t r a n s f o r m e d_.

3. Replaced gps_location into gps_lat, gps_long and gps_alt.

As Chris mentioned, the new CSV files (from Ford 2019) contain gps_lat, gps_long and gps_alt, instead of gps_location. I wasn't sure if it's possible to have an if statement inside the constructor to check if the CSV file has gps_location OR gps_lat/long/alt. So I decided that it might be better to simply have gps_lat/long/alt as the instance variables. Since gps_location will no longer be a column in future CSV files, then the chunks of code written for handling gps_location will be eventually redundant. 

This will simply mean that the script will only work for CSV files AFTER Ford 2019. (If the columns are the same)

That's all the key changes I've made. Let me know if you have a second opinion/idea for the key changes I've made and other feedback. Thanks!

khanguslee:
> 1. Moved the write_to_output_file outside the constructor, previously it was the last line in `__init__`.
> My initial idea for this script was to automate everything for the user, but this is better because it allows flexibility in how the user wants the data to be _t r a n s f o r m e d_.

- Your logic makes sense if we wanted to write a script that only does one thing. But if we want to think about it long term, we can reuse your class in different ways!

> 1. Replaced gps_location into gps_lat, gps_long and gps_alt.
> 
> This will simply mean that the script will only work for CSV files AFTER Ford 2019. (If the columns are the same)

- Yeah oops, this is my fault. Perhaps we can create some sort of converter that just converts V2 to V3 logs one day..

khanguslee (dismissed):
LGTM. Only gripe I have is docstrings! But we can sort that out when we have a standard we want in the DAS repo.

hallgchris:
> Yeah oops, this is my fault. Perhaps we can create some sort of converter that just converts V2 to V3 logs one day...

Agreed we should make a converter between these two formats, but this script isn't for V3 logs is it? Cos V3 logs have different columns depending on the sensor module

utils/das_data_sorter.py
```diff
@@ -122,20 +122,22 @@ def average_data(self, data:pd.Series) -> pd.Series:
...
+        for index_array in self.indexes:
+            if 1 in data[index_array].values:
+                new_gps_data.append(1)
+            else:
+                new_gps_data.append(0)
         
-    #     return new_gps_data
+        return new_gps_data
 
     def location_data(self, data:pd.Series) -> pd.Series:
```
hallgchris: Now that we're not dealing with the `gps_location` column, we can actually average the data, as it's no longer stored as a string!

hallgchris (changes requested):
Lucky last thing. All good after this!

khanguslee:
Thoughts on having `poetry` inside the `utils` folder? It seems like `poetry` is located inside the `DAS` folder in https://github.com/monash-human-power/data-acquisition-system/pull/53. Should we just have a single poetry file at the root of the directory? Not sure what would be the best solution. @Blake-Haydon @hallgchris @nix-chu @angustrau.

hallgchris:
I reckon just have one single poetry file at the root per project. Unless one option is clearly better than the other (and do speak up if one is!), the simplest solution is always the best of two equal options

nix-chu:
Ah, I did this just to try installing poetry myself. I don't mind taking it out of utils folder if having a single poetry file works just fine :)

khanguslee:
@nix-chu Just leave your poetry file then. Would be cool if poetry had something like `yarn workspaces`.

nix-chu:
@hallgchris do you mean have one in my project and another in @Blake-Haydon's etc? or just one for the entire repository?

hallgchris:
I was saying one for the entire repo, but thinking more about it, I'm happy to leave it as is. @Blake-Haydon's one is for stuff to be ran on the raspberry pi, yours is for stuff to run later on our computers, so I can see why they should be kept separate.

khanguslee:
Just wanted to note that, if you do a `poetry install` to install your dependencies for this project, it will create a `.egg-info` folder. See https://github.com/python-poetry/poetry/issues/800 for more information. To remove this, you will need to install libraries by doing `poetry install --no-root`. I think this will be fixed when we move to a single poetry file at the root of this project.
Screen Shot 2020-05-25 at 3 21 36 pm

khanguslee (approved):
Output from your script! (Don't forget these on your PR next time pls!)
Screen Shot 2020-05-25 at 3 38 55 pm

Such a useful script!

hallgchris (approved):
Looks great!

PR #51, opened 2020-04-20
Implements smoothing technique into converted data ver2

nix-chu:
## Description
Refer to the ver1 of this script for Purpose of this script, How Moving Mean/Median Works, and Advantages & Disadvantages of Smoothing Techniques. The difference between ver1 and ver 2 is; when converting from milliseconds to seconds/minutes, this ver2 script already applies the smoothing technique. No extra averaging is done. 

Advantages of version 2 compared to version 1:
- utilises less code lines
- runs faster
- possibly easier to understand

Disadvantages of version 2 compared to version 1:
- Doesn't seem to have changed anything, using either technique

**only approve ONE of the PRs**

## Screenshots
(Using Mean Smoothing technique)
Left: Normal data
Right: Mean Smoothed data
image

(Using Median Smoothing technique)
Left: Normal data
Right: Median Smoothed data
image

## Steps to Test
1. In the utils folder, transfer one of the data files from the DAS into this folder
2. In the terminal, run `python das_data_sorter --file (file name) --output (output name) --units (seconds/minutes) --smooth (mean/median)`

### DIFF
--- a/utils/Pipfile
+++ b/utils/Pipfile
@@ -9,6 +9,7 @@ verify_ssl = true
 pandas = "*"
 argparse = "*"
 math = "*"
+numpy = "*"
 
 [requires]
 python_version = "3.8"
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -1,19 +1,23 @@
 import pandas as pd 
 import argparse 
 from math import ceil
+from numpy import median, mean
 
 # averages data within the seconds, according to seconds array 
-def avg_Data(array):
-    avg_array = []
+def avg_Data(array, technique):
+    new_data = []
 
-    for index_array in new_time:
-        total = 0
-        for index in index_array:
-            total += array[index]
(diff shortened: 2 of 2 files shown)


nix-chu:
As I kept writing this code, I'm more aware of how messy my script is. I apologise for that. As Angus L suggested in April 20 meeting, I will transform all of this into a (hopefully) readable Class in Python.

khanguslee (commented):
OHHHH, okay. I see the difference now. There's a big difference between #50 and this PR. For now, I prefer version 1 (#50) since it averages the time within each second. The newly sampled data should then be filtered using mean or median smoothing. 

If we really wanted to change the way we sort the times into seconds/minutes, we would actually apply the smoothing algorithm across the whole data FIRST. We would then sort that new data into seconds/minutes. Otherwise if there's a value jump between 999ms and 1001ms, this script does not filter that out.

PR #52, opened 2020-04-27
Add wireless sensor python framework

emilytrau:
## Description

This PR adds some framework code to kickstart the python version of the wireless module firmware. From here we can implement MQTT and the different sensor types (the DHT sensor is implemented as an example).  This is moved over from monash-human-power/wireless-modules#4

## Screenshots

N/A

## Steps to Test

Follow the getting started steps to upload the code to an ESP32.

### DIFF
--- a/WirelessModules/.editorconfig
+++ b/WirelessModules/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/WirelessModules/LICENSE
+++ b/WirelessModules/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Monash-Human-Power
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/WirelessModules/Pipfile
+++ b/WirelessModules/Pipfile
@@ -1,0 +1,14 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+pylint = "*"
+
+[packages]
+micropy-cli = "*"
+rshell = "*"
+
+[requires]
+python_version = "3.8"
--- a/WirelessModules/README.md
+++ b/WirelessModules/README.md
@@ -1,0 +1,50 @@
+# DAS Wireless Sensor Modules
+
+Wireless sensor firmware based on MicroPython.
+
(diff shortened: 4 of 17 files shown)


emilytrau:
@khanguslee I think enforcing docstrings is a good idea. I'll implement automated linting in a later PR

khanguslee:
> @khanguslee I think enforcing docstrings is a good idea. I'll implement automated linting in a later PR

I like this! Thanks!

khanguslee:
Just to reiterate, I would like for everyone to approve this (since it's the foundation of wireless modules onto the das repo) after reading and understanding the code of course.

khanguslee (approved):
LGTM

WirelessModules/.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
```
hallgchris, line 6: This config demands 2 space indents, but files seem to have 4 space indents.
(I personally would prefer 4, and that's what we have elsewhere for python)

WirelessModules/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
hallgchris, line 1: @khanguslee You said earlier that we should make some more repos private, including this one. Do we need a license if it's private?

WirelessModules/Pipfile
```diff
@@ -0,0 +1,14 @@
...
+
+[dev-packages]
+pylint = "*"
+
+[packages]
+micropy-cli = "*"
+rshell = "*"
+
+[requires]
+python_version = "3.8"
```
hallgchris, line 14: Love how ESP32s run a newer version of python than the raspberry pi

WirelessModules/README.md
```diff
@@ -0,0 +1,50 @@
...
+# Install the required tools
+pipenv install
+pipenv shell
+
+# Configure VS-Code auto-completion for micropython
+micropy stubs add esp32-micropython-1.12.0
+micropy
+```
+
+You will need to create a local version of `config.py` (`cp src/config.example.py src/config.py`) and change some settings.
```
hallgchris, line 21: Not all of us use unix (unfortunately) so perhaps suggest windows users `copy` instead (I think that's right?)

WirelessModules/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Flashing MicroPython to the ESP32
+
+The ESP32 needs to be flashed with the base micropython firmware. This only needs to be done once. See the [MicroPython ESP32 Getting Started guide](https://docs.micropython.org/en/latest/esp32/tutorial/intro.html) for detailed instructions.
+
+## Uploading Code
+
+```
+# Change this to the serial port of the ESP32
+./upload.sh --port /dev/ttyUSB0
```
hallgchris, line 37: As @khanguslee said in the old PR, this won't work on windows. Perhaps we need a batch file, or at least some instructions for windows.

lakshjaisinghani (approved):
Looks good to me

WirelessModules/src/wireless_module.py
```diff
@@ -0,0 +1,13 @@
...
+        super().__init__()
+        self.sensors = []
+
+    def add(self, sensor):
+        self.sensors.append(sensor)
+    
+    def read_sensors(self):
+        readings = []
+        for sensor in self.sensors:
+            readings += sensor.read()
```
hallgchris, line 12: Had no idea you could append elements to an array with tuples like that. Super cool!

WirelessModules/src/boot.py
```diff
@@ -0,0 +1 @@
+# boot.py - - runs on boot-up
```
hallgchris, line 1: Don't have an ESP32 on me to test for myself - if this file runs on boot, `main.py` won't ever be called, right? Or I'm missing something

hallgchris (changes requested):
I can't comment on the files as they're empty so doing it here, but is there a reason we have `requirements.txt` with pipenv?

Got a few questions and suggestions, but looks really great. Seeing this code has me really excited that we're using micropython for this.

WirelessModules/Pipfile
khanguslee: Whoever has the RPI's, maybe try: https://installvirtual.com/how-to-install-python-3-8-on-raspberry-pi-raspbian/

WirelessModules/LICENSE
khanguslee: That's a valid question actualy. There is merit in adding a license to a private repo since it then provides some sort of license on the code that is put on the code in this repository. Since MHP doesn't require us to write any NDA or anything.

Instead of an MIT license, are there any licenses that we should look into @hallgchris? See https://choosealicense.com/

Blake-Haydon (approved):
Just checked the upload and everything works fine! It works awesome!

WirelessModules/src/boot.py
emilytrau: This is an extra script that gets run before main.py. We can ignore it for now

hallgchris:
The middle wireless module has been implemented in micropython with the fantastic work of @kdav108, incorporating many of the ideas of this PR. See #61

PR #53, opened 2020-05-04
Fix DAS structure

Blake-Haydon:
## Description
This turned out to be a bigger PR than I wanted by it hopefully fixes a few issues with the logging as well as the file structure of the DAS repo.

**Key changes**
- Fixed logger not logging
- Renamed many files so that they make sense and moved to directories that also make sense
- Added poetry (Removed pipenv)

## Screenshots

image

## Steps to Test

***This is taken from the new read me in the DAS directory***

## **Basic Setup and Usage**
These steps assume that you are in the data-acquisition-system root directory.
1. `cd DAS`
2. `poetry install`
3. `poetry shell`
4. `cd ../`
5. To test this PR run `python -m DAS.mqtt_wireless_logger` in one terminal and `python -m DAS.tests.fake_module_tester --time 10 --rate 2` in another.
6. There should be saved csv files stored in the `DAS/csv_data` after about 10 secs. 

Below is just more info if you need it

-----

## **mqtt_wireless_logger**
Once started this script will log any data coming in on the MQTT module data and battery channels. The saved csv files will be stored in the `DAS/csv_data`. 

**Usage**: From the data-acquisition-system root directory run `python -m DAS.mqtt_wireless_logger --host 192.168.1.100`. This will connect to the MQTT broker set to 192.168.1.100 and run a logger indefinitely. If no host is specified it will default to `localhost`.

<br>

## **fake_module_tester**
This script mocks module data over MQTT similar to the real sensors on V3.

**Usage**: From the data-acquisition-system root directory run `python -m DAS.tests.fake_module_tester --host 192.168.1.100 --time 10 --rate 2 --id 3`. This will run a fake test with the MQTT broker on 192.168.1.100 for 10 seconds, pinging out data 2 times per second only for module 3.

<br>


AntPlusSensor/README.md
```diff
@@ -1,3 +1,9 @@
+# Ant Plus Sensor
+
+### DAS.js
+node.js program we run to communicate to the Teensy that is connected to the Raspberry Pi via serial communication.
```
hallgchris: This is outdated, `DAS.js` no longer has anything to do with serial communication with a Teensy.

DAS/README.md
```diff
@@ -0,0 +1,40 @@
...
+<br>
+
+## **fake_module_tester**
+This script mocks module data over MQTT similar to the real sensors on V3.
+
+**Usage**: From the data-acquisition-system root directory run `python -m DAS.tests.fake_module_tester --host 192.168.1.100 --time 10 --rate 2 --id 3`. This will run a fake test with the MQTT broker on 192.168.1.100 for 10 seconds, pinging out data 2 times per second only for module 3.
+
+<br>
+
+## **serial_test**
```
hallgchris: Don't need serial test anymore :(

DAS/mqtt_wireless_logger.py
```diff
@@ -1,21 +1,31 @@
...
 import os
 import pandas as pd
-from DataToTempCSV import DataToTempCSV
 from datetime import datetime
 import argparse
-import enum_topics
+import glob
+import re
+from DAS.utils.DataToTempCSV import DataToTempCSV 
+from DAS.utils import TopicsEnum
```
hallgchris: Is there a way to not need to prefix imports with `DAS.`?

DAS/tests/mqtt_test.py
```diff
@@ -5,24 +5,34 @@
...
+                    help="Starts replaying from a specified time (in seconds)")
 
 
 def start_publishing(client, args):
     print("start")
     client.publish("power_model/start")
     client.publish("start")
 
-    send_data_func = lambda data: client.publish("data", data)
+    def send_data_func(data): return client.publish("data", data)
```
hallgchris, line 35: If we're switching from a lambda to a function (any particular reason?) probably put the return statement on a new line.

README.md
```diff
@@ -13,40 +13,6 @@ This repository contains all the code related to Monash Human Power's Data Acqui
...
-Once started this script will log any data coming in on the MQTT module data and battery channels. The saved csv files will be in the format `module_id + _D%d-%m-%d_T%H-%M-%S + .csv`, where the date and time is the time when the start topic gets data sent on it.
-
-#### Usage: Just start the script and data will be auto logged
-- General use with the default MQTT broker set to localhost
-  - `python3 mqtt_wireless_log.py`
-- Same logger with the MQTT broker set to 192.168.1.100
-  - `python3 mqtt_wireless_log.py --host 192.168.1.100`
-
-
 ## Teensy
```
hallgchris: This bit can be deleted too.

hallgchris (changes requested):
`.eslintrc`, `.eslintignore`, `.prettierrc` and `.gitignore` (all in `DAS` folder) are all files related to the javascript part of the das, and should be moved to `AntPlusSensor`.

Also, `serial_test.py` can be deleted rather than moved - don't need it for V3.

Not sure why github doesn't let me comment on moved files which aren't modified :sweat_smile: 

Wondering what other people think - upper or lower case folder names? I'm inclined to go with snake case, for consistency, but happy to go with this too.

DAS/README.md
khanguslee: Should we reference our V2 tag on the README so we don't forget about these code snippets? https://github.com/monash-human-power/data-acquisition-system/tree/v2

AntPlusSensor/README.md
khanguslee: Adding on to this, `DAS.js` is the only way we can communicate to ant+ devices. https://github.com/Loghorn/ant-plus

Side note: We should rename this file to avoid confusion later on..

DAS/README.md
```diff
@@ -0,0 +1,40 @@
...
+These steps assume that you are in the data-acquisition-system root directory.
+1. `cd DAS`
+2. `poetry install`
+3. `poetry shell`
+4. `cd ../`
+5. Run any DAS command, eg. `python -m DAS.mqtt_wireless_logger`
+
+<br>
+
+## **mqtt_wireless_logger**
```
khanguslee: - Add file extension in to headers?
- Add a link to the actual file as well.

DAS/mqtt_wireless_logger.py
khanguslee: See https://alex.dzyoba.com/blog/python-import/ on importing python files. 

Doesn't look like we can remove the DAS bit though.

DAS/mqtt_wireless_logger.py
```diff
@@ -1,21 +1,31 @@
 import paho.mqtt.client as mqtt
 import os
 import pandas as pd
-from DataToTempCSV import DataToTempCSV
 from datetime import datetime
 import argparse
-import enum_topics
+import glob
+import re
+from DAS.utils.DataToTempCSV import DataToTempCSV
```
khanguslee: Can you make it so that we write `from DAS.utils import DataToTempCSV`. You will need to do something similar to this i think: https://stackoverflow.com/questions/38927979/default-export-in-python-3

DAS/tests/fake_module_tester.py
```diff
@@ -2,8 +2,8 @@
 import argparse
 import json
 import paho.mqtt.client as mqtt
-from MockSensor import MockSensor
-import enum_topics
+from DAS.utils.MockSensor import MockSensor
```
khanguslee: Same with this, is there anyway we can just do `from DAS.utils import MockSensor`?

khanguslee (changes requested):
Logging works! But it leaves an empty `.~temps` folder behind after logging. Can we remove this or have the temp files stored temporarily inside `csv_data`.
Screen Shot 2020-05-14 at 3 23 17 pm

Your move breaks `DAS.js` btw. Please change line 5 to:
``` js
const RollingAverage = require("./average");
```

khanguslee:
`mqtt_test.py` does not work! Can you double check that all other scripts work?
Screen Shot 2020-05-14 at 5 34 33 pm

AntPlusSensor/README.md
Blake-Haydon: Cool, I wasn't too sure about the serial code so I just kept it just in case.

DAS/tests/mqtt_test.py
Blake-Haydon: I did not do that, it appears as if the atom linter did it automatically. I'll change it back to a lambda as its much nicer

DAS/README.md
Blake-Haydon: Yeah just did that, once it get squashed into the main branch it will work

khanguslee:
Seems like there's a conflict with the README!

Blake-Haydon:
I have fixed many of the pressing issues. However some things I have chosen to keep the same:
- `DAS.utils.MockSensor import MockSensor` was not changed to `DAS.utils import MockSensor`. This was because when trying to implement it for all there was a circular reference. It will be fixed in another PR, but it will require more refactoring.
- `.~temps` folder was kept as removing it is very difficult as for some reason the temp csvs do not show up when using glob

khanguslee:
> * `DAS.utils.MockSensor import MockSensor` was not changed to `DAS.utils import MockSensor`. This was because when trying to implement it for all there was a circular reference. It will be fixed in another PR, but it will require more refactoring.

Thanks for trying! I'm glad you worked out that it might take longer to complete and it warranted another PR. I would say it's a nice to have, but not crucial if it takes too long to do! Hopefully it was a good exercise in learning Python modules!

> * .~temps folder was kept as removing it is very difficult as for some reason the temp csvs do not show up when using glob

Maybe we can figure this out in another PR?

khanguslee (approved):
LGTM!

ant_plus_sensor/ant_plus_logger.js
```diff
@@ -7,7 +7,7 @@ const RollingAverage = require('./utils/average');
 const argumentParser = new ArgumentParser({
   version: '1.0.0',
   addHelp: true,
-  description: 'MHP DAS',
+  description: 'Ant-Plus MQTT Logger',
```
hallgchris, line 10: Don't really care for this PR but so you know, Ant-Plus is usually stylised as Ant+

hallgchris (approved):
Looks great! Will be really nice to work with this folder structure that actually makes sense

PR #54, opened 2020-06-01
Add Sample Micropython

Blake-Haydon:
## Description

A sample MQTT test that runs on ESP32 using micro-python. 

## Steps to Test

In order to run this code you need an ESP32:
1. Use a tool such as ampy or rshell to copy all of the files to the ESP32
2. Restart the ESP32 and you should see an output when using mosquitto_sub or the DAS data logger script 

If you want to turn on and off the LED remotely send data to the v3/wireless-module/{MODULE_ID}/LED topic with either  "on"  or "off"


wireless_modules/mqtt_test/config.example.py
```diff
@@ -0,0 +1,4 @@
+ESSID = "<WIFI NETWORK NAME HERE>"
```
khanguslee, line 3: Can these values be read into from a `.env` file instead? If we can, we can have a `config.py` script that just reads the values there.

wireless_modules/mqtt_test/config.example.py
```diff
@@ -0,0 +1,4 @@
...
+PASSWORD = "<WIFI PASSWORD HERE>"
+
+MQTT_BROKER = "192.168.1.111"
```
khanguslee: Any reason for this IP address?

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,27 @@
+from config import ESSID
+from config import PASSWORD
+
+def connect(eesid, password):
+    import network
```
khanguslee: Is this some micropython thing where you import inside a function?

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,27 @@
...
+    sta_if = network.WLAN(network.STA_IF)
```
khanguslee: I had to google what `STA_IF` meant. Let's rename `sta_if` to `station_interface` or `network_interface` instead? Thoughts?

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,27 @@
...
+    if not sta_if.isconnected():
+        print('connecting to network...')
+        sta_if.active(True)
+        sta_if.connect(eesid, password)
+        while not sta_if.isconnected():
+            pass
```
khanguslee, line 20: What did you want this to do? A comment might help with what you intend for this whole loop to do.

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,27 @@
...
+
+    print('network config:', sta_if.ifconfig())
+
+def no_debug():
```
khanguslee: What does this function do?

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,27 @@
...
+    import esp
+    esp.osdebug(None)
+
+# Bootup and connect to wifi credentials stored in config.py
+no_debug()
+connect(ESSID, PASSWORD)
+
+# Turn on the built in LED when booted (blue light)
+from machine import Pin
```
khanguslee: Any possibility we can put all imports at the top of the python file?

wireless_modules/mqtt_test/fake.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def battery(self):
+        """ Decrease battery % by 1 and convert to JSON string """
+        self.battery_pc -= 1
+        battery_dict = {"percentage": self.battery_pc}
+        return ujson.dumps(battery_dict)
+
+
+
+class Sensor:
```
khanguslee: If we wanted this class to be proper, we should have a `Sensor` interface in which a `TemperatureSensor` will implement. Currently your class describes a sensor with many types of sensors which doesn't make sense. This is probably unnecessary because it's fake so a better implementation could be a `FakeSensor` class where you include the type as a parameter into the constructor. Example: `FakeSensor("temperature")`. 

An even more elegant solution would be to have a generic `FakeSensor` class where you input a string AND `Value` class. Example:`FakeSensor("temperature", RandomValue)` or `FakeSensor("accelerometer", XYZValue)`.

Feel free to choose! There's just a lot of duplicate code and don't think we should ignore it even if it is fake. This is probably what we will end up using too!

wireless_modules/mqtt_test/main.py
```diff
@@ -0,0 +1,100 @@
+import time
+import machine
+from machine import Pin
+import ubinascii
+from config import MQTT_BROKER
+from umqttsimple import MQTTClient
+import fake
+
+# Define module number and MQTT topics
+MODULE_NUM = "1"
```
khanguslee, line 20: Can ESP32 micropython use `.env` files?

wireless_modules/mqtt_test/main.py
```diff
@@ -0,0 +1,100 @@
...
+SUB_TOPICS = [SUB_LED]
+
+# Define built in LED
+builtin_LED = Pin(2, Pin.OUT)
+
+# Set up MQTT client by generating a client_id
+client_id = ubinascii.hexlify(machine.unique_id())
+
+# MQTT callbacks
+def sub_cb(topic, msg):
```
khanguslee, line 35: Not really fussed if this whole file isn't a class yet.

wireless_modules/mqtt_test/umqttsimple.py
```diff
@@ -0,0 +1,207 @@
...
+    import usocket as socket
+except:
+    import socket
+import ustruct as struct
+from ubinascii import hexlify
+
+class MQTTException(Exception):
+    pass
+
+class MQTTClient:
```
khanguslee: Is this just the paho-mqtt library code that's been replicated here? One question I have is, what do we do when the paho-mqtt library code is updated? Do we need to update this manually? Can we automate this somehow?

khanguslee (changes requested):
Just got a few questions about micropython. Some small variable name changes too. Really exciting to learn about micropython in general! Good work on this!

wireless_modules/mqtt_test/config.example.py
Blake-Haydon: It was just the one I was using at home. What is the DAS IP? I'll change it to the DAS one

wireless_modules/mqtt_test/boot.py
Blake-Haydon: Yeah it speeds up the code so that you only fetch things when you need them

Blake-Haydon: All of the examples use this. I'm not sure how much of a difference it makes though.

wireless_modules/mqtt_test/boot.py
Blake-Haydon: It allows for the code to be uploaded and run. If you don't disable it, you may not be able to use ampy or rshell to upload your code to the board

Blake-Haydon: I'll add more comments to the scripts as it is a bit confusing

wireless_modules/mqtt_test/boot.py
Blake-Haydon: It just waits for the internet to connect before moving on.

Blake-Haydon: A comment would be handy as it is a bit confusing

wireless_modules/mqtt_test/umqttsimple.py
Blake-Haydon: There is no official 'paho-mqtt' library for MicroPython. I assume because of the limited space requirements and the fact that paho is meant to have all of the features of MQTT. The MicroPython `umqttsimple.py` script is a slimmed down client with all the key features we need. 

About updating, I'm really not sure how this would work as I literally just copied from the repo to get this to work. Because the MQTT standard does not change often, I can't imagine we would have to update the script regularly. 

I am very keen for input however, as I feel like it is a bit of a hack.

[This it the umqttsimple repo if you want to check it out](https://github.com/RuiSantosdotme/ESP-MicroPython/blob/master/code/MQTT/umqttsimple.py)

[Alternative MQTT implementation](https://github.com/micropython/micropython-lib/tree/master/umqtt.simple)

wireless_modules/mqtt_test/config.example.py
Blake-Haydon: Just looked it up and it doesn't seem like you are able to use .env files without having to read it just like a normal file.

wireless_modules/mqtt_test/boot.py
Blake-Haydon: I think it is a good idea! `sta_if` is confusing when you first read it and `station_interface` is much more clear. I like it!

wireless_modules/mqtt_test/fake.py
Blake-Haydon: Yeah 100%. I kinda hacked it together so I will try and implement it correctly

wireless_modules/mqtt_test/main.py
Blake-Haydon: Doesn't look like it

wireless_modules/mqtt_test/umqttsimple.py
Blake-Haydon: Okay this is actually a really easy fix.

We can just use upip, which is micro-pip. It will just check whether the latest version of MQTT is installed and install the latest version automatically.

wireless_modules/mqtt_test/main.py
Blake-Haydon: Yeah this is also a bit hacky, but it works for now

Blake-Haydon:
Things that have been improved:
- Huge redundancy in fake.py has been eliminated!
- No more hardcoded MQTT client, using upip now
- Added better comments

wireless_modules/mqtt_test/config.example.py
hallgchris: For V2 we used `192.168.100.100`. I'd really like to use a thing called mDNS though, cos then we could use something like `das.local`. But that IP should be good for now

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,34 @@
...
+    print('network config:', station_interface.ifconfig())
+
+# Allows the code to be uploaded with tools such as rshell / ampy
+esp.osdebug(None)
+
+# Connect to WiFi using credentials stored in config.py
+connect(config.ESSID, config.PASSWORD)
+
+# Turn on the built in LED when booted and connected to WiFi (blue light)
+builtin_LED = Pin(2, Pin.OUT)
```
hallgchris: Is the builtin LED always on pin 2, or could it vary between development boards? (Does micropython have some constant like arduino's `LED_BUILTIN` we could use?)

wireless_modules/mqtt_test/main.py
```diff
@@ -0,0 +1,108 @@
...
+import upip
+upip.install("micropython-umqtt.simple2")
+from umqtt.simple import MQTTClient
+
+
+# Define module number and MQTT topics
+MODULE_NUM = "1"
+FAKE_MODULE = fake.module(MODULE_NUM)
+
+PUB_DATA = b'/v3/wireless-module/{}/data'.format(MODULE_NUM)
```
hallgchris, line 23: Does whatever version of micropython we use support f-strings?

wireless_modules/mqtt_test/main.py
```diff
@@ -0,0 +1,108 @@
...
+        builtin_LED.on()
+
+    if topic == SUB_LED and msg == b'off':
+        print('ESP received: LED OFF')
+        builtin_LED.off()
+
+def connect_and_subscribe(client_id, mqtt_server, sub_topics):
+    client = MQTTClient(client_id, mqtt_server)
+    client.set_callback(sub_cb)
+    client.connect()
```
hallgchris, line 49: What happens if connection fails?

wireless_modules/mqtt_test/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+PUB_DATA = b'/v3/wireless-module/{}/data'.format(MODULE_NUM)
+PUB_BATTERY = b'/v3/wireless-module/{}/battery'.format(MODULE_NUM)
+PUB_LOW_BATTERY = b'/v3/wireless-module/{}/low-battery'.format(MODULE_NUM)
+
+SUB_LED = b'/v3/wireless-module/{}/LED'.format(MODULE_NUM)
+SUB_TOPICS = [SUB_LED]
+
+# Define built in LED
+builtin_LED = Pin(2, Pin.OUT)
```
hallgchris: Is it possible to use the same `builtin_LED` definition here and in `boot.py`? Maybe it could be stored in config

hallgchris (changes requested):
This looks really good! Super excited to work with micropython eventually. I need to get myself an ESP32.

Just got a few questions, mainly cos I'm unfamiliar with micropython. Pretty much good to go though, if think all this is good then happy to approve in its current state

Edit: actually, Angus has some good points below, worth doing haha

wireless_modules/mqtt_test/fake.py
```diff
@@ -0,0 +1,90 @@
...
+        """ Decrease battery % by 1 and convert to JSON string """
+        
+        self.battery_pc -= 1
+        battery_dict = {"percentage": self.battery_pc}
+        return ujson.dumps(battery_dict)
+
+
+
+class FakeSensor:
+    def __init__(self, sensor_name):
```
khanguslee, line 49: This looks a lot better now! It could be improved but I prefer this over a large if statement any day!

wireless_modules/mqtt_test/boot.py
```diff
@@ -0,0 +1,34 @@
+import config
```
khanguslee, line 1: Any chance we can catch an error if `config.py` does not exist and/or cannot be found? Only asking because it will happen and people will ask you how to fix it!

I would re-raise the exception with an additional message stating to create a `config.py` script.

ALSO, can you update `README.md` with setup and how to run this?

wireless_modules/mqtt_test/fake.py
```diff
@@ -0,0 +1,90 @@
...
+                fake_gps_data = FakeSensor(sensor_name).value_GPS()
+                data_dict["sensors"].append(fake_gps_data)
+
+            elif sensor_name == "accelerometer" or sensor_name == "gyroscope":
+                fake_xyz_data = FakeSensor(sensor_name).value_xyz()
+                data_dict["sensors"].append(fake_xyz_data)
+
+            else:
+                fake_single_data = FakeSensor(sensor_name).value_single()
+                data_dict["sensors"].append(fake_single_data)
```
khanguslee: I would bring out `data_dict["sensors"].append(fake_single_data)` from every if statement by just making the variable name of the fake sensor more generic!

Outside if statement:
`data_dict["sensors"].append(fake_sensor)`

Once you bring this out, can you see how this part of this code actually starts to look like a potential `FakeSensorFactory` class. See SimpleFactory here: https://refactoring.guru/design-patterns/factory-comparison . Not needed for this PR (factory class), but just wanted you to see how you sometimes implement design patterns without realising that you are already using one!

khanguslee (changes requested):
Code looking a bit more cleaner now! There's a few things that I missed that I think would be worth it for the long run.

- Exception message that says you're missing a `config.py` file
- `README.md` on how to setup these files + installation. Maybe put this inside the `wireless-modules/` folder?
- One code cleanup thingo.

Thanks for the quick fix up from the last review!

wireless_modules/mqtt_test/boot.py
khanguslee: I was waiting for you to write this @nix-chu

hallgchris: Just a thought, should this be a json or .env config file rather than an actual python file? Not sure if that'll work in micropython though

nix-chu: > I was waiting for you to write this @nix-chu

@khanguslee config file or a readme?

khanguslee: I believe @Blake-Haydon said `.env` files don't work in micropython. IDM a JSON file as well?

@nix-chu Oh I meant the suggestion since you said the same thing in my PR. 😄

wireless_modules/mqtt_test/boot.py
Blake-Haydon: It doesn't seem to have a builtin led variable. I have moved it to the LED initiation to the config file top make it simpler.

wireless_modules/mqtt_test/main.py
Blake-Haydon: No I dont believe so as I had trouble getting them to work.

wireless_modules/mqtt_test/main.py
Blake-Haydon: It tries again. Ill comment it to make it easier to read

Blake-Haydon:
Only some small fixes this time:
+ All declarations of LED's have been moved to the config
+ Small bugs have been fixed
+ I added _fun_ status lights that help with debugging

wireless_modules/mqtt_test/boot.py
Blake-Haydon: I think actual python is good, as it is handy to initiate sensors and LEDs in this file.

Blake-Haydon: > Any chance we can catch an error if config.py does not exist and/or cannot be found? Only asking because it will happen and people will ask you how to fix it!

I have added @angustrau build script so hopefully that doesn't happen

hallgchris: > I have added @angustrau build script so hopefully that doesn't happen

Not quite sure what you're referring to there. Just want to point out how this could be done quite easily if you're unsure about it:
```python
try:
    import config
except ImportError:
    print("error message here")
```

wireless_modules/mqtt_test/boot.py
```diff
@@ -25,10 +23,12 @@ def connect(eesid, password):
 
 # Allows the code to be uploaded with tools such as rshell / ampy
 esp.osdebug(None)
+config.red_LED.on()
```
hallgchris, line 26: While these LEDs certainly look brilliant, I'm unsure about their relevance in some sample code. I feel like this code should be able to be run by anyone with an ESP32 with minimal setup, but this would require them to set up a circuit exactly like you have. I kinda feel like they perhaps shouldn't be included. Thoughts @khanguslee ?

hallgchris (changes requested):
Really close! The ESP32 with the blinky lights looks super cool.

wireless_modules/mqtt_test/boot.py
khanguslee: Yeah I agree. Happy to use the built-in LED for sample code that we see here. Alternatively, you can just print out information instead.

hallgchris: Yeah to be clear I'm fine with the built-in LED too, since all ESP32s have that without any setup

wireless_modules/mqtt_test/boot.py
Blake-Haydon: Oops my bad, I read it wrong, will add in the try catch now.

wireless_modules/mqtt_test/boot.py
Blake-Haydon: Yeah that's fair enough, I just reasoned that if the code was to be run on a board without LEDs attached it simply would do nothing.

Blake-Haydon:
Did only small fixes:
- Added try statement for the config import
- Add comment about LED's

hallgchris (approved):
I would still lean towards not having them at all, but since this is just a demo it's probably fine. Great job on this PR!

khanguslee (approved):
LGTM!

PR #55, opened 2020-06-06
Implement Smoothing Techniques on DAS Data

nix-chu:
**Updated as of 14/06, read for new screenshots and steps to run**
## Description
The following changes implements 2 smoothing techniques into `das_data_sorter.py` to filter out unwanted data and fluctuations. The smoothing techniques are called 'Moving Mean' and 'Moving Median'. The output will be used for BOOST's Power Model. 

Will write documentation for how Smoothing Technique works after exam period, or after this PR. However, if you want to know how it works now:
- [Video explaining Smoothing Time Series Data](https://www.youtube.com/watch?v=U0dF8uJs3mE)
- [Timestamp for Moving Mean for Odd values of N](https://youtu.be/U0dF8uJs3mE?t=52)
- [Timestamp for Moving Mean for Even values of N](https://youtu.be/U0dF8uJs3mE?t=396)
- [Timestamp for Moving Median for Odd values of N](https://youtu.be/U0dF8uJs3mE?t=676)
- [Timestamp for Moving Median for Even values of N](https://youtu.be/U0dF8uJs3mE?t=1084)

To sum it up, Moving Mean and Moving Median basically 'smooth' the data points. Starting from the first data point, it will take N consecutive number of data points and find the mean/median between them, then it will move up on data point and take the next N consecutive number of data points. By doing this, this will give the average of the entire data and remove noise. An example of 3-point (odd) Moving Mean is shown below:
image

Having a higher value for N means more smoother data points and less fluctuations, however, this will result in having less data points overall. 

Refer down below for photos of how even-numbered N works.

## Screenshots
Left: Regular filtered data with **no** smoothing
Right: Smoothed and filtered data with smoothing
CSV file used: data_291.csv

Mean smoothing using 23 data points:
image

Median smoothing using 23 data points:
image

## Steps to Test
For converting from milliseconds to seconds of data **(no smoothing)**:
1. In the terminal, run `python das_data_sorter.py --input <file name> --output <output name> --unit <seconds/minutes>`
Unit has a default of seconds
2. When finished, it should print 'Success!' into the terminal

For applying smoothing techniques:
1. In the terminal, run `python das_data_sorter.py --input <file name> --output <output name> --unit <seconds/minutes> --smooth <mean/median> -n <number of data points>`
2. When finished, it should print 'Success!' into the terminal

### DIFF
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -6,9 +6,10 @@ from numpy import ceil, median
 parser = ArgumentParser()
 parser.add_argument("--input", help="Reads the inputted CSV file to filter", action="store", required=True)
 parser.add_argument("--output", help="Writes the filtered data onto a new CSV file under this name", action="store", required=True)
-parser.add_argument("--unit", help="Specify time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
+parser.add_argument("--unit", help="Specifies time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
                     choices=["seconds", "s", "minutes", "m"], action="store")
 parser.add_argument("--smooth", help="Smooths data points using N-point mean or median smoothing", choices=["mean", "median"], action="store")
+parser.add_argument("--n", help="Specifies number of data points taken for smoothing", action="store", type=int)
 args = parser.parse_args()
 
 class DasSort:
@@ -136,6 +137,7 @@ class DasSort:
         return new_gps_data
     
     def mean_smooth_caller(self, n:int=3) -> None:
+        '''Calls __mean_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         
@@ -145,6 +147,11 @@ class DasSort:
                 self.__dict__[variable] = self.__mean_smooth(self.__dict__[variable], n)
 
     def __mean_smooth(self, data:pd.Series, n:int) -> pd.Series:
+        '''Returns an array of smoothed data based on number of data points taken to smooth. 
+        
+        For an odd number N, the data points are simply averaged. 
+        Whereas for an even number N, the data points are averaged, then centered. This is because the data point will misalign with integer
+        numbers of time if not done.'''
         smooth_data_array = [] 
         if n % 2 != 0: # Mean Smoothing for odd number N
             for i in range(len(data) - n + 1):
@@ -170,6 +177,7 @@ class DasSort:
         return smooth_data_array 
     
     def median_smooth_caller(self, n:int=3) -> None:
+        '''Calls __median_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         
@@ -179,6 +187,11 @@ class DasSort:
                 self.__dict__[variable] = self.__median_smooth(self.__dict__[variable], n)
     
     def __median_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
+        '''Returns an array of smoothed data based on number of data points taken to smooth. 
+        
+        For an odd number N, the data points are simply averaged. 
+        Whereas for an even number N, the data points are averaged, then centered. This is because the data point will misalign with integer
+        numbers of time if not done.'''
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         smooth_data_array = [] 
@@ -237,7 +250,7 @@ if __name__ == '__main__':
     data = pd.read_csv(args.input) # Loads and reads CSV file
     das_sort = DasSort(data, args.unit) # Filters data in CSV file
     if args.smooth == "mean": # Applies smoothing technique, when provided
-        das_sort.mean_smooth_caller()
+        das_sort.mean_smooth_caller(args.n)
     elif args.smooth == "median":
-        das_sort.median_smooth_caller()
+        das_sort.median_smooth_caller(args.n)
     das_sort.write_to_output_file(args.output) # Write filtered data into new CSV file
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -136,54 +136,58 @@ class DasSort:
 
         return new_gps_data
     
-    def smooths_data(self, data:pd.Series, technique:str, n:int=3) -> pd.Series:
+    def mean_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         smooth_data_array = [] 
 
-        if technique == "mean" and n % 2 != 0: # Mean Smoothing for odd number N
+        if n % 2 != 0: # Mean Smoothing for odd number N
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+n]
                 new_data_point = self.mean(data_points)
                 new_data_point = round(new_data_point, ndigits=2)
                 smooth_data_array.append(new_data_point)
-        elif technique == "mean" and n % 2 == 0: # Mean Smoothing for even number N
+        else: # Mean Smoothing for even number N
             temp_array = []
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+N]
                 new_data_point = self.mean(data_points)
-                new_data_point = round(new_data_point, ndigits=2)
                 temp_array.append(new_data_point)
             
-            # centres the data by averaging adjacent data points
+            # an extra step to centre the data by averaging adjacent data points
             for j in range(len(temp_array) - 1): 
                 first = temp_array[j]
                 second = temp_array[j+1]
                 new_data_point = self.mean([first,second])
                 new_data_point = round(new_data_point, ndigits=2)
                 smooth_data_array.append(new_data_point)
-        elif technique == "median" and n % 2 != 0: # Median Smoothing for odd number N
+        
+        return smooth_data_array 
+    
+    def median_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        smooth_data_array = [] 
+
+        if n % 2 != 0: # Median Smoothing for odd number N
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+n]
                 new_data_point = median(data_points) # round not required, since it will always be the middle number
                 smooth_data_array.append(new_data_point)
-        elif technique == "median" and n % 2 == 0:
+        else: # Median Smoothing for even number N
             temp_array = []
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+n]
                 new_data_point = median(data_points)
-                new_data_point = round(new_data_point, ndigits=2)
                 temp_array.append(new_data_point)
             
-            # centres the data by averaging adjacent data points
+            # an extra step to centre the data by averaging adjacent data points
             for j in range(len(temp_array) - 1):
                 first = temp_array[j]
                 second = temp_array[j+1]
                 new_data_point = median([first,second])
                 new_data_point = round(new_data_point, ndigits=2)
                 smooth_data_array.append(new_data_point)
-        else:
-            raise Exception("Code broken. Choices should be limited to [median,mean] and invalid N values should be caught beforehand.")
 
         return smooth_data_array 
     
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -193,5 +193,5 @@ if __name__ == '__main__':
     data = pd.read_csv(args.input) # Loads and reads CSV file
     das_sort = DasSort(data, args.unit) # Filters data in CSV file
     if args.smooth: # Applies smoothing technique, when provided
-        das_sort.smooth_caller(args.n, args.smooth)
+        das_sort.smooth(args.n, args.smooth)
     das_sort.write_to_output_file(args.output) # Write filtered data into new CSV file
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -53,7 +53,7 @@ class DasSort:
         else:
             raise ValueError("Argument only accepts either seconds/s or minutes/m")
         
-        # sets the groups of indexes to use for averaging data, once time has been sorted
+        # sorts into groups of indexes to average columns in self.data
         self.indexes = self.__group_index(new_time)
 
         return range(1, len(self.indexes) + 1)
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -138,14 +138,17 @@ class DasSort:
 
         return new_gps_data
     
-    def mean_smooth_caller(self, n:int) -> None:
+    def smooth_caller(self, n:int, technique:str) -> None:
         '''Calls __mean_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
         if n < 3 or n > len(self.indexes):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         
-        # passes instance variables as an argument in smoothing function
-        for variable in self.data:
-            self.data[variable] = self.__mean_smooth(self.data[variable], n)
+        if technique == "mean":
+            for variable in self.data:
+                self.data[variable] = self.__mean_smooth(self.data[variable], n)
+        elif technique == "median":
+            for variable in self.data:
+                self.data[variable] = self.__median_smooth(self.data[variable], n)
 
     def __mean_smooth(self, data:pd.Series, n:int) -> pd.Series:
         '''Returns an array of smoothed data based on number of data points taken to smooth. 
@@ -154,7 +157,7 @@ class DasSort:
         Whereas for an even number N, the data points are averaged, then centered. This is because the data point will misalign with integer
         numbers of time if not done.'''
         smooth_data_array = [] 
-        
+
         if n % 2 != 0: # Mean Smoothing for odd number N
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+n]
@@ -177,15 +180,6 @@ class DasSort:
                 smooth_data_array.append(new_data_point)
         
         return smooth_data_array 
-    
-    def median_smooth_caller(self, n:int) -> None:
-        '''Calls __median_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
-        if n < 3 or n > len(self.indexes):
-            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
-        
-        # passes instance variables as an argument in smoothing function
-        for variable in self.data:
-            self.data[variable] = self.__median_smooth(self.data[variable], n)
     
     def __median_smooth(self, data:pd.Series, n:int) -> pd.Series:
         '''Returns an array of smoothed data based on number of data points taken to smooth. 
@@ -226,8 +220,6 @@ class DasSort:
 if __name__ == '__main__':
     data = pd.read_csv(args.input) # Loads and reads CSV file
     das_sort = DasSort(data, args.unit) # Filters data in CSV file
-    if args.smooth == "mean": # Applies smoothing technique, when provided
-        das_sort.mean_smooth_caller(args.n)
-    elif args.smooth == "median":
-        das_sort.median_smooth_caller(args.n)
+    if args.smooth: # Applies smoothing technique, when provided
+        das_sort.smooth_caller(args.n, args.smooth)
     das_sort.write_to_output_file(args.output) # Write filtered data into new CSV file
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -88,7 +88,6 @@ class DasSort:
         '''
         total = 0
         length = len(data_array)
-
         for number in data_array:
             if number > 0 or number < 0:
                 # 'number' value is valid
@@ -136,11 +135,17 @@ class DasSort:
 
         return new_gps_data
     
-    def mean_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
+    def mean_smooth_caller(self, n:int=3) -> None:
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
-        smooth_data_array = [] 
+        
+        # passes instance variables as an argument in smoothing function
+        for variable in self.__dict__:
+            if variable != "indexes": # excludes self.indexes
+                self.__dict__[variable] = self.__mean_smooth(self.__dict__[variable], n)
 
+    def __mean_smooth(self, data:pd.Series, n:int) -> pd.Series:
+        smooth_data_array = [] 
         if n % 2 != 0: # Mean Smoothing for odd number N
             for i in range(len(data) - n + 1):
                 data_points = data[i:i+n]
@@ -164,7 +169,15 @@ class DasSort:
         
         return smooth_data_array 
     
-    def median_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
+    def median_smooth_caller(self, n:int=3) -> None:
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        
+        # passes instance variables as an argument in smoothing function
+        for variable in self.__dict__:
+            self.__dict__[variable] = self.__median_smooth(self.__dict__[variable], n)
+    
+    def __median_smooth(self, data:pd.Series, n:int=3) -> pd.Series:
         if n < 3 or n > len(data):
             raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
         smooth_data_array = [] 
@@ -222,4 +235,8 @@ class DasSort:
 if __name__ == '__main__':
     data = pd.read_csv(args.input) # Loads and reads CSV file
     das_sort = DasSort(data, args.unit) # Filters data in CSV file
+    if args.smooth == "mean": # Applies smoothing technique, when provided
+        das_sort.mean_smooth_caller()
+    elif args.smooth == "median":
+        das_sort.median_smooth_caller()
     das_sort.write_to_output_file(args.output) # Write filtered data into new CSV file
--- a/utils/das_data_sorter.py
+++ b/utils/das_data_sorter.py
@@ -161,6 +161,7 @@ class DasSort:
                 first = temp_array[j]
                 second = temp_array[j+1]
                 new_data_point = self.mean([first,second])
+                new_data_point = round(new_data_point, ndigits=2)
                 smooth_data_array.append(new_data_point)
         elif technique == "median" and n % 2 != 0: # Median Smoothing for odd number N
             for i in range(len(data) - N + 1):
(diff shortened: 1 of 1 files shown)


khanguslee:
Did I just get forced to watch a Further Maths video?

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
         
         return new_gps_data
     
+    def smooths_data(self, data:pd.Series, technique:str, n:int=3) -> pd.Series:
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        smooth_data_array = [] 
+
+        if technique == "mean" and n % 2 != 0: # Mean Smoothing for odd number N
+            for i in range(len(data) - N + 1):
```
hallgchris: Where is `N` defined? Also, in line with `snake_case` conventions, it's probably better to call the variable `n`.

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
         
         return new_gps_data
     
+    def smooths_data(self, data:pd.Series, technique:str, n:int=3) -> pd.Series:
```
hallgchris: This function is quite large with lots of if statements, which can be difficult to maintain - I'd probably make two smaller functions, one to handle mean and one to handle median, which this function may call.

utils/das_data_sorter.py
```diff
@@ -1,13 +1,14 @@
...
-from numpy import ceil
+from numpy import ceil, median
 
 # accepts terminal arguments
 parser = ArgumentParser()
 parser.add_argument("--input", help="Reads the inputted CSV file to filter", action="store", required=True)
 parser.add_argument("--output", help="Writes the filtered data onto a new CSV file under this name", action="store", required=True)
 parser.add_argument("--unit", help="Specify time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
                     choices=["seconds", "s", "minutes", "m"], action="store")
+parser.add_argument("--smooth", help="Smooths data points using 3-point mean or median smoothing", choices=["mean", "median"], action="store")
```
hallgchris: You say 3-point mean/median here - do you plan be able to do other numbers of points?

hallgchris (changes requested):
Just had a quick look through - I'll do a proper review when you're done coding and I can run it.
I think it could be okay to call your function from `__main__`, but you'll have to change it so that it updates the class' state rather than returning a new smoothened array, so that the smoothened data is written when `write_to_output_file` is called.

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
         
         return new_gps_data
     
+    def smooths_data(self, data:pd.Series, technique:str, n:int=3) -> pd.Series:
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        smooth_data_array = [] 
+
+        if technique == "mean" and n % 2 != 0: # Mean Smoothing for odd number N
```
khanguslee: I would make a `mean_smoothing` function that encapsulates the if statement for the `technique == "mean"` pathway. I don't think `smooths_data` should care if your mean or median smoothing technique needs to run an odd or even pathway, it should just call `mean_smoothing` or `median_smoothing`.

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+                new_data_point = round(new_data_point, ndigits=2)
+                temp_array.append(new_data_point)
+            
+            # centres the data by averaging adjacent data points
+            for j in range(len(temp_array) - 1): 
+                first = temp_array[j]
+                second = temp_array[j+1]
+                new_data_point = self.mean([first,second])
+                smooth_data_array.append(new_data_point)
+        elif technique == "median" and n % 2 != 0: # Median Smoothing for odd number N
```
khanguslee: `median_smoothing` method instead. See above comment.

utils/das_data_sorter.py
khanguslee: Where did `N` come from?

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+
+        if technique == "mean" and n % 2 != 0: # Mean Smoothing for odd number N
+            for i in range(len(data) - N + 1):
+                data_points = data[i:i+N]
+                new_data_point = self.mean(data_points)
+                new_data_point = round(new_data_point, ndigits=2)
+                smooth_data_array.append(new_data_point)
+        elif technique == "mean" and n % 2 == 0: # Mean Smoothing for even number N
+            temp_array = []
+            for i in range(len(data) - N + 1):
```
khanguslee: This seems like duplicated code as the odd pathway? Maybe we just need an if statement that checks if it's even so that it can run the for loop below?

utils/das_data_sorter.py
```diff
@@ -136,6 +137,53 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+                new_data_point = self.mean([first,second])
+                smooth_data_array.append(new_data_point)
+        elif technique == "median" and n % 2 != 0: # Median Smoothing for odd number N
+            for i in range(len(data) - N + 1):
+                data_points = data[i:i+N]
+                new_data_point = median(data_points) # round not required, since it will always be the middle number
+                smooth_data_array.append(new_data_point)
+        elif technique == "median" and n % 2 == 0:
+            temp_array = []
+            for i in range(len(data) - N + 1):
```
khanguslee: Also feels like duplicate code as above. I think I would rather apply a `round` method (even though it does nothing) than to have duplicate code!

khanguslee (changes requested):
What data are you intending to smooth btw? That will affect my answer as to how you would run this code.

utils/das_data_sorter.py
nix-chu: Oh whoops sorry, that's a typo. It is meant to be lowercase n

utils/das_data_sorter.py
nix-chu: Oh yes, I was planning to have an input for other numbers of points. That's another typo 😅 thanks for pointing them out

utils/das_data_sorter.py
nix-chu: Typo 😅 haha thanks for pointing it out

nix-chu:
> Did I just get forced to watch a Further Maths video?

@khanguslee Ay but if I didn't do Further Maths in VCE, I would not have known how to smooth data.

nix-chu:
@hallgchris Good recommendation, I'll have to pass each instance variable as an argument for the smoothing method. I was thinking of creating an array of instance variables, but that sounds very tacky? Like as in, I'm sure there is a better way of doing that. Thoughts?

nix-chu:
@khanguslee all data columns have to be smoothed the same way, in order to have each data point aligned with a corresponding time. So to answer your question, it's either all the columns or none of them. The user has to choose.

hallgchris:
> @hallgchris Good recommendation, I'll have to pass each instance variable as an argument for the smoothing method. I was thinking of creating an array of instance variables, but that sounds very tacky? Like as in, I'm sure there is a better way of doing that. Thoughts?

Real sorry, not quite sure what you're referring to. What instance variables?

utils/das_data_sorter.py
nix-chu: Will definitely separate mean_smoothing and median_smoothing! Forgot, we mentioned that last time too. 

Although, I think an odd-numbered N pathway and an even-numbered N pathway should remain separate. It's because the even-numbered pathway has an extra step where the averaged data points is 'centered'. In the video, for a 3-point mean smoothing, the averaged data point was already centered with a corresponding time. Whereas for a 4-point mean smoothing, the average data point was in-between numbers, so an extra step of 'centering' was done.

I actually had the same idea about putting an if statement to check for an even number N, but it wouldn't work out (unless you think of something, please let me know). 

Basically, if I stored the averaged data points into `smooth_data_array`, then I can take the values from the same array to do centering and then update the values in the same array (without overwriting something needed). Sounds like it works. 

However, the problem is that 'centering' reduces the number of data points, so you won't have the same number of data points. This means that when centering is finished, you will have leftover data points that are not part of the final converted data. This leftover data points could be a random spike or part of the original data. Because of this issue and the extra steps, I was thinking of separating them, and storing values in two separate arrays for even-numbered N.

If there's time tomorrow, I can demonstrate with a diagram during/after the meeting, if unclear.

nix-chu:
> Real sorry, not quite sure what you're referring to. What instance variables?

Time, GPS, temperature, etc. Basically whatever is in the `__init__`

khanguslee:
> @khanguslee all data columns have to be smoothed the same way, in order to have each data point aligned with a corresponding time. So to answer your question, it's either all the columns or none of them. The user has to choose.

What I was referring to was, are you looking to replace the `average_data` function so that within each time interval, you use the median value instead? OR is this for applying a smoothing algorithm after you've averaged things into second/minutes? (I think it's the latter right?)

nix-chu:
> > @khanguslee all data columns have to be smoothed the same way, in order to have each data point aligned with a corresponding time. So to answer your question, it's either all the columns or none of them. The user has to choose.
> 
> What I was referring to was, are you looking to replace the `average_data` function so that within each time interval, you use the median value instead? OR is this for applying a smoothing algorithm after you've averaged things into second/minutes? (I think it's the latter right?)

Yup the latter. Applying a smoothing algorithm after averaging data into seconds/minutes.

khanguslee:
> > > @khanguslee all data columns have to be smoothed the same way, in order to have each data point aligned with a corresponding time. So to answer your question, it's either all the columns or none of them. The user has to choose.
> > 
> > 
> > What I was referring to was, are you looking to replace the `average_data` function so that within each time interval, you use the median value instead? OR is this for applying a smoothing algorithm after you've averaged things into second/minutes? (I think it's the latter right?)
> 
> Yup the latter. Applying a smoothing algorithm after averaging data into seconds/minutes.

In that case, a user should be able to have smoothing off, mean smoothing applied or median smoothing applied (not both at the same time). I'm currently leaning towards having it being called in the `if __name__ = "__main__":` section. I view  your class as some sort of a pipeline where the dev can clearly see the steps that your class takes and what operations are applied to the data.

I hope that answers your question. @hallgchris thoughts about this idea?

hallgchris:
>  @hallgchris thoughts about this idea?

That sounds good to me!

nix-chu:
Honestly, so proud of this PR right now. I can feel myself working better at 2am not gonna lie. Will list major changes below:

1. Separated mean smooth and median smooth methods. 

I have separated the mean smooth and median smooth methods which will hopefully make the script easier to read. I have kept the if statement check for odd or even numbers of N, specifically because the smoothing techniques for odd numbered N differs for even numbered N. 

Smoothing technique for even numbered N requires an extra step of centering, because without this step, the data points would be misaligned similar to shown below. 
image
By doing an extra step of centering, the smoothing technique more accurately represents the actual data point at that timestamp:
image
As you can see from above, a lot of data points were removed. However, this is the consequence of increasing the value N, regardless if N was odd or even. Our files have plenty of data anyways, so too much data won't be lost for an appropriate value of N. 

2. Created method callers 

To follow Pythonic standards, I've created separate method callers. I've done this so that instance variables would not need to be called outside the class and make it possible to iterate through each instance variable (excluding indexes which is not in the output file). However, on second thought, instead of having two separate method callers, we could have one. Let me know your opinions. 

Learned that I could iterate class & instance variables by referencing `__dict__`. More info: https://stackoverflow.com/questions/25150955/python-iterating-through-object-attributes 
Code written by top-voted seems to be outdated. Tried finding a more recent post but couldn't. Hope this is something new for you guys to learn. 

3. Integrated smoothing functions which now are being called in `if __name__ == '__main__'` HOORAY 🎉

Thank you for providing your inputs, it was very helpful. 

As I've updated in the PR description, to run this script to its fullest potential: `python das_data_sorter.py --input <file name> --output <output name> --unit <seconds/minutes> --smooth <mean/median> -n <number of data points>`

nix-chu:
As a somewhat important side-note, in terms of using an if statement to check for odd or even numbered N. I have considered simply using one if statement for the block of code responsible for centering. The issue with this is that, although I can overwrite the elements in the `smooth_data_array`, I would end up having incorrect data at the end of the array.

As you've seen from the two photos in the comment above, due to smoothing+centering, there is a lower number of total data points than original. Therefore, even though I can overwrite the elements in the `smooth_data_array`, I'll have irrelevant data points trailing at the end of the array and I won't know how many data points are considered irrelevant. Which is why I have kept the initial if statement check. 

Hope my two comments are easy to understand, because I am really tired right now. Let me know if you have questions tomorrow.

utils/das_data_sorter.py
```diff
@@ -1,13 +1,15 @@
...
 
 # accepts terminal arguments
 parser = ArgumentParser()
 parser.add_argument("--input", help="Reads the inputted CSV file to filter", action="store", required=True)
 parser.add_argument("--output", help="Writes the filtered data onto a new CSV file under this name", action="store", required=True)
-parser.add_argument("--unit", help="Specify time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
+parser.add_argument("--unit", help="Specifies time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
                     choices=["seconds", "s", "minutes", "m"], action="store")
+parser.add_argument("--smooth", help="Smooths data points using N-point mean or median smoothing", choices=["mean", "median"], action="store")
+parser.add_argument("--n", help="Specifies number of data points taken for smoothing", default=3, action="store", type=int)
```
hallgchris: Usually for single-letter flags like this we only use a single dash, i.e. `-n`.

In fact, we often use single letter flags as shortcuts for longer ones. If you felt like it, you could add `-i` and `-o` as shortcuts for input and output :)

utils/das_data_sorter.py
```diff
@@ -128,14 +127,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
     
+    def mean_smooth_caller(self, n:int) -> None:
+        '''Calls __mean_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        
+        # passes instance variables as an argument in smoothing function
+        for variable in self.__dict__:
+            if variable != "indexes": # excludes self.indexes
+                self.__dict__[variable] = self.__mean_smooth(self.__dict__[variable], n)
```
hallgchris: While iterating through all of the variables of an object like this a super cool idea, I can see some potential issues with it down the track. Whenever we want to add some new variable to the class, lets say we want to retain the file name as an example, we're going to have to add another check that `variable != "filename"` here, and everywhere else that we iterate through `self.__dict__`. It seems managable here with just one exception like you've got here (`indices`) but every single time someone adds a variable, they're going to have to change this code, which they mightn't even be aware of. This smells of questionable programming practice, simply adding a variable to a class shouldn't affect the way unrelated code runs.

Let's think of how we could get around this. Currently, you're treating the class as a dictionary, with fields as keys and lists of data as values. Rather than using the class as a dictionary, you could simply contain all this data in a dedicated dictionary! This also means you'll combine many instance variables of this class into a single variable, which is great because this class certainly has a few (also considered non-ideal)

utils/das_data_sorter.py
```diff
@@ -128,14 +127,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+        '''Calls __median_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
+        if n < 3 or n > len(data):
+            raise ValueError("Number of smoothing points must be at least 3 and less than the length of the data set to perform smoothing.")
+        
+        # passes instance variables as an argument in smoothing function
+        for variable in self.__dict__:
+            if variable != "indexes": # excludes self.indexes
+                self.__dict__[variable] = self.__median_smooth(self.__dict__[variable], n)
+    
+    def __median_smooth(self, data:pd.Series, n:int) -> pd.Series:
```
hallgchris: As far as I can see, `__median_smooth` is the same as `__mean_smooth` except one calls `median` and the other calls `self.mean`. Instead of having two functions, you could just have a single function which takes some function `average_func` as a parameter. (yes, you can pass functions as variables! super cool) You could then call it with either
```python
self.__smoothen("gps_speed", n, median)
```
or
```python
self.__smoothen("gps_speed", n, self.mean)
```
and then inside `__smoothen`,
```python
def __mean_smooth(self, data: pd.Series, n: int, average_func) -> pd.Series:
    ...
    new_data_point = average_func(data_points)
    ...
```

hallgchris: Note I'm going with the technical meaning of the work average, where both median and mean are a type of average (https://en.wikipedia.org/wiki/Average)

utils/das_data_sorter.py
```diff
@@ -128,14 +127,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
         for index_array in self.indexes:
             if 1 in data[index_array].values:
                 new_gps_data.append(1)
             else:
                 new_gps_data.append(0)
-        
+
         return new_gps_data
     
+    def mean_smooth_caller(self, n:int) -> None:
```
hallgchris: When writing a class' public interface, always think about how it would feel to call stuff from outside of the class. Does it make sense to someone who doesn't know about the inside workings of `DasSort` to call something called `mean_smooth_caller`? To them, `_caller` doesn't mean anything. I think it would be fine if this were just called `mean_smooth`.

utils/das_data_sorter.py
```diff
@@ -167,4 +249,8 @@ def write_to_output_file(self, file_output:str) -> None:
 if __name__ == '__main__':
     data = pd.read_csv(args.input) # Loads and reads CSV file
     das_sort = DasSort(data, args.unit) # Filters data in CSV file
+    if args.smooth == "mean": # Applies smoothing technique, when provided
+        das_sort.mean_smooth_caller(args.n)
+    elif args.smooth == "median":
+        das_sort.median_smooth_caller(args.n)
```
hallgchris: You asked about whether having two separate functions or a single function to do the smoothening would be better. I think what you've got already is best in this case!

If we had a single function, we'd have to also pass in what type of smoothening we wanted, probably as a string. We'd then have to worry about checking that the string is a valid option, and stuff like case sensitivity in a class that ideally shouldn't be concerned about such stuff. By only having two choices (by calling one of either function), it's impossible to tell `DasSort` to do an invalid smoothening technique.

utils/das_data_sorter.py
hallgchris: Looking back at my older comments I realise it was me that told them to split them up... Sorry about that! I guess now with them split up it became easier to see how similar the code was and how it could be combined but more concisely with less repetition than before. That's part of programming I guess!

hallgchris (changes requested):
Great stuff! Awesome to hear that you're so proud your work :)

Got some technical comments here... Feel free to ask any questions about them

utils/das_data_sorter.py
nix-chu: A very clever idea in the long run! 

Will have to make big changes throughout the entire code, but I think it will be worth it

utils/das_data_sorter.py
nix-chu: I think I might leave the two `__mean_smooth` and `__median_smooth` functions separate, just for readability and its easier to maintain. However, I could simply have one caller function and pass `average_func` as the argument. Would this ruin the readability in `if __name__ == '__main__'`? 

I'll could commit this change and then see people's opinions on this idea.

nix-chu:
```python
def smooth_caller(self, data:pd.Series, n:int, average_func) -> pd.Series:
    if average_func == "mean":
        self.__mean_smooth(data, n)
    elif average_func == "median":
        self.__median_smooth(data, n)
    else:
        raise ValueError
...
if __name__ == '__main__':
    if args.smooth:
        smooth_caller(args.n, args.smooth)
```

So this will keep the two smoothing functions separate, but the caller function is the same. I know you've said to get rid of the 'caller' part, I just used it as an example for now. Thoughts?

nix-chu:
Some new changes:
1. Argument for N number of data points has changed to `-n`
2. All data columns have been stored into one dictionary and assigned as `self.data`
3. (Messed up the commit message for this one and I don't know how to revert it back) Combined the two smoothing callers into one singular caller `smooth_caller`

Previously, with two separate smoothing callers, the if statement at the end of the script looked like this
image

Advantage of using the previous code is that its more readable, more detailed for future Electrical team to follow how the script works when they scroll all the way to the end.

At the moment, I have changed it so that there is only one smoothing call
image

Advantage of using this is that it's more concise (and maybe, in some other way, easier to read)

utils/das_data_sorter.py
```diff
@@ -128,43 +129,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
         for index_array in self.indexes:
             if 1 in data[index_array].values:
                 new_gps_data.append(1)
             else:
                 new_gps_data.append(0)
-        
+
         return new_gps_data
     
+    def smooth_caller(self, n:int, technique:str) -> None:
```
hallgchris: Again, `_caller` is probably not necessary in this function name. I think you might have accidentally readded it when you combined the two functions into one haha

utils/das_data_sorter.py
```diff
@@ -128,43 +129,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+        '''Calls __mean_smooth function to smooth the data in instance variables, without requiring to refer to the variables themselves.'''
```
hallgchris: Documentation needs updating - no longer only calls `__mean_smooth`

utils/das_data_sorter.py
hallgchris: While I understand what you're saying about increased readability, I don't think it'd be easier to maintain. Interested to know why you disagree!

As I see it, we've got two almost identical functions. If we need to change how one of them works in the future, we'll almost certainly want to also change the other one. This means we'd have to remember to change code in two locations for the one change, which makes it harder to maintain.

utils/das_data_sorter.py
```diff
@@ -128,43 +129,97 @@ def gps_data(self, data:pd.Series) -> pd.Series:
...
+            for i in range(len(data) - n + 1):
+                data_points = data[i:i+n]
+                new_data_point = median(data_points)
+                temp_array.append(new_data_point)
+            
+            # an extra step to centre the data by averaging adjacent data points
+            for j in range(len(temp_array) - 1):
+                first = temp_array[j]
+                second = temp_array[j+1]
+                new_data_point = median([first,second])
```
hallgchris: I think finding the mean here would be more clear than finding the median, even though it gives the same result for a two element array. Just cos when I think of what operation is used to find the value exactly half way between two others, my mind always goes to mean rather than median.

hallgchris (changes requested):
Sorry for all the changes. Happy to hear your point of view if you disagree with anything!

Kinda a general comment for a lot of the documentation - I think it would be good if we're careful about where we use the words average, mean and median. Average is like a general term that could refer to either, so any time you're refering to only mean (for example) it's best to use the word mean, not average.

utils/das_data_sorter.py
nix-chu: Oh right, it is harder to maintain. I was just giving a word vomit of words thatll make me sound like I know what im doing lol, don't mind me

utils/das_data_sorter.py
nix-chu: Should this be changed to `smooth`?

utils/das_data_sorter.py
nix-chu: I don't mind doing this, I was just thinking whether changing this to `self.mean` for a Median technique would be confusing for future Electrical team who are learning about 'Moving Average' techniques for the first time. Would it be confusing or am I overthinking this?

@khanguslee @minh-tn would like your opinions about this as well

utils/das_data_sorter.py
```diff
@@ -1,39 +1,43 @@
...
-        self.gy = self.average_data(file_input["gY"])
-        self.gz = self.average_data(file_input["gZ"])
-        self.thermoc = self.average_data(file_input["thermoC"])
-        self.thermof = self.average_data(file_input["thermoF"])
-        self.pot = self.average_data(file_input["pot"])
-        self.cadence = self.average_data(file_input["cadence"])
-        self.power = self.average_data(file_input["power"])
-        self.reed_velocity = self.average_data(file_input["reed_velocity"])
-        self.reed_distance = self.average_data(file_input["reed_distance"])
+        self.data = {
```
khanguslee, line 19: This is definitely a good candidate to refactor later on! But no need to worry about it for now!

utils/das_data_sorter.py
```diff
@@ -1,39 +1,43 @@
...
+from numpy import ceil, median
 
 # accepts terminal arguments
 parser = ArgumentParser()
-parser.add_argument("--input", help="Reads the inputted CSV file to filter", action="store", required=True)
-parser.add_argument("--output", help="Writes the filtered data onto a new CSV file under this name", action="store", required=True)
-parser.add_argument("--unit", help="Specify time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds", 
+parser.add_argument("-i", "--input", help="Reads the inputted CSV file to filter", action="store", required=True)
+parser.add_argument("-o", "--output", help="Writes the filtered data onto a new CSV file under this name", action="store", required=True)
+parser.add_argument("--unit", help="Specifies time units [seconds, s, or minutes, m]. Default is in seconds.", default="seconds",
```
khanguslee: I know this was added in a different PR, but you don't need to specify what units are valid `[seconds, s, or minutes, m]` because the `--help` will do that for you.
Screen Shot 2020-06-20 at 12 06 53 am

utils/das_data_sorter.py
khanguslee: I would also remove the `_caller` bit to the method name. As you use more and more libraries in various languages, you will learn how to call your method names. As a developer, it would be a bit weird for me to add on the `_caller` bit because I don't care if it's a function that calls another function. I just want it to smooth my data! So yes, `smooth` would be a good method name for this!

utils/das_data_sorter.py
khanguslee: Gonna have to agree with @hallgchris and say that I would want to have one single function that takes in the smoothing function as a parameter to this function. This is because the code is actually duplicated and like stated before, will be harder to maintain. This should also resolve the issue below of applying median vs mean to a two element array because we just apply  the stated smoothing function.

khanguslee (changes requested):
Thanks for the relatively small PR! Just a few things that will make your code more maintainable.

utils/das_data_sorter.py
hallgchris: I think for a person learning about how the maths works, the most confusing point would be why we even do this second round of averaging to center the data. If they've figured that out, I think using mean wouldn't be any more confusing than median. Besides, I reckon using mean better describes what we're trying to do, so it could be less confusing!

nix-chu:
New changes:

- Combined mean smooth and median smooth functions into one smooth function

Reused my `self.mean` function, since I already had it written. I thought, why not make more use out of it instead of importing another mean function. (Also found out that you cannot pass a string as a callable function, so I had to write `self.mean` and `median`)

- Changed group indexes function into an internal function

Figured that this should be an internal function, because grouping the indexes should always run after time has been converted, since a 'group of indexes' represent the indexes of the data in a single time interval. `group_index` has no purpose being called outside the class imo. 

On that note, I've always had a problem with how this algorithm was run, not sure whether if it's an efficiency problem or design problem. It would feel reassuring if you guys had a second look at this method because something about it bothers me, but not sure what.

hallgchris (approved):
Looks good! Thanks for making those changes.

> I've always had a problem with how this algorithm was run, not sure whether if it's an efficiency problem or design problem

It's definitely possible to compute a rolling mean more efficiently, and possibly also for a rolling median. (For mean, you could have a queue of a fixed size, and a sum variable. Every time you look at a new element, add its value to the sum, push it to the queue, then pop an element from the queue and subtract that from the sum. The rolling average is that sum divided by the size of the queue). But I wouldn't worry about that for a few reasons:

- The increased efficiency would be traded off with the increased complexity of the code to handle the cases we currently handle, like when a value is 0, when a value is n/a, the centering you perform, etc.
- Maximum efficiency isn't really a priority here, as this isn't a mission-critical script that's running continuously on the bike while riding or anything like that, it's only ran as a once-off for analysis.

On the other hand, optimising code for speed can be fun! So feel free to if you feel like it, but I wouldn't expect you too do it. I'm sure you're looking forward to starting something new soon

khanguslee (approved):
LGTM! Hope this whole process didn't scare you away from coding! As you write more code and create pull requests, the pull request comments will eventually decrease!

Good work!

PR #57, opened 2020-07-21
Fix DAS module structure

hallgchris:
## Description

I'd noticed that we could only run DAS scripts from the root directory of this repo, and were forced to use the `python -m` syntax. To fix this, I've moved all the python code down a directory relative to the `pyproject.toml`. I realised this may have been the issue after @khanguslee's work on monash-human-power/common.

I've also simplified some imports e.g. `from DAS.utils.DataToTempCSV import DataToTempCSV` to ` from das.utils import DataToTempCSV`

While I was at it, I renamed the package from `DAS` to `das` to conform to [PEP 8 conventions](https://www.python.org/dev/peps/pep-0008/#package-and-module-name).

## Screenshots

N/A, just moving files about

## Steps to Test

`poetry init`, `poetry shell`, then check that all your favorite scripts run using either (e.g.) `python das/tests/mqtt_test.py` or `python -m das.tests.mqtt_test`.


khanguslee (approved):
YAY! I think we all have a better understanding of this now. I would like @Blake-Haydon to approve this before we merge this in. But this gets my LGTM.

Should we move the `utils` folder into a `utils` repo? We could add some old stuff on there too. Obviously another PR.

nix-chu (commented):
Still getting an import error, maybe this is just a me problem? Not sure whats wrong

image

`python -m -m das.tests.mqtt_test` works tho

hallgchris:
@nix-chu did you run `poetry install` and `poetry shell`?

nix-chu (approved):
Ohhhh okay I see it now. That's my bad. All goods here! 👍

image

Blake-Haydon (approved):
This looks amazing! Glad you were able to fix the structure.

PR #58, opened 2020-08-04
Add MQTT Client class 

kdav108:
## Description

This pull request adds an additional folder in data-acquisition-system/wireless_modules called middle_module which contains files that help the ESP32 connect to the WIFI and a client Class file which can be used to help connect, send and receive messages over the MQTT messaging protocol. This files is independent and so can be re-used by anyone wanting to use MQTT. 
The folder also contains a test file and a main.py which is still in the works to read sensor data for the middle wireless module.

## Screenshots

The test.py output (on the left) and the main.py output (on the right):
Test main

## Steps to Test

These are included in the README in the new folder, but I have it restated here again:
1. Ensure your ESP32 board is connected to the WIFI, to do this you can use the boot.py file and a locally created config.py and connect it (this would also In the process confirm that boot.py works)
2.  Open the test.py file on a IDE on the local computer, ensure the paho-mqtt package for python is installed
3. Place the main.py and MQTT_Class_client.py files onto the ESP32 board
4. Run the main.py file and without wasting too much time run the test.py file
5. You should see the test.py output messages sent from main.py on the console. This would confirm that both main.py and MQTT_Client_class.py works (main.py uses all the methods in the Client class)


wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,51 @@
...
+
+client = mqtt.Client("my_id")
+
+MODULE_NUM = "2"
+START_TOPIC = '/v3/wireless-module/{}/start'.format(MODULE_NUM)
+STOP_TOPIC = '/v3/wireless-module/{}/stop'.format(MODULE_NUM)
+SUB_DATA_TOPIC = '/v3/wireless-module/{}/data'.format(MODULE_NUM)
+
+# Broker domain name or IP address
+BROKER = "5.196.95.208"
```
emilytrau: Could you make this an argument? Same as with our other test scripts

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,51 @@
+# Author: Kunj
+# Last Modified: 3/08/2020
+import paho.mqtt.client as mqtt
+import time
+
+client = mqtt.Client("my_id")
+
+MODULE_NUM = "2"
+START_TOPIC = '/v3/wireless-module/{}/start'.format(MODULE_NUM)
```
emilytrau: Try [f-strings](https://realpython.com/python-f-strings/#simple-syntax) for cleaner syntax

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,51 @@
+# Author: Kunj
+# Last Modified: 3/08/2020
```
emilytrau: Git will keep track of who and when modified this code. I don't think we need to explicitly include this

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,58 @@
...
+    attempts = 5
+
+    # Create station interface
+    station_interface = network.WLAN(network.STA_IF)
+
+    # Activate the interface
+    station_interface.active(True)
+
+    if station_interface.isconnected():
+        print('network config:', station_interface.ifconfig())
```
emilytrau: Make sure that you consistently use either  `'` or `"` throughout the code where possible

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,102 @@
...
+upip.install("micropython-umqtt.simple2")
+try:
+    from umqtt.simple import MQTTClient
+except FileNotFoundError:
+    print("Error in importing library")
+
+
+class Client:
+    def __init__(self, client_id, mqtt_broker):
+        """
```
emilytrau: Great documentation! :+1:

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,102 @@
...
+        This functions takes care of all of the formatting and publishes 'data' on the given 'topic'. It also checks for
+        any pending message, but this should not be relied upon solely - use the check_for_message() method externally
+        when an incoming message is to be checked for/read.
+        :param data: A string of data to be sent
+        :param topic: A string representing the topic to send 'data' to.
+        :return: Nothing
+        """
+        try:
+            # convert the data to bytes literal so that MQTT can read it
```
emilytrau: You've over-commented this a bit. Comments should describe _why_ you're doing something rather than _what_ your doing. For small-to-medium sized functions the overall function comment is usually sufficient.

emilytrau (changes requested):
Congrats on submitting your first PR! :tada:

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: You mean no need for comments within the function if I have the overall function comment?

wireless_modules/middle_module/test.py
kdav108: Oh damn never knew this exited 😂

wireless_modules/middle_module/MQTT_Client_class.py
emilytrau: Right. You only need to comment more on particularly complex blocks where you feel the need to explain your line of thought and why you did something that way.

wireless_modules/middle_module/MQTT_Client_class.py
khanguslee: I really like this documentation style. Is this some sort of docstring convention?

wireless_modules/middle_module/MQTT_Client_class.py
khanguslee: Just adding on, the code that was written in this function was already self explanatory. The docstring for this function also explained what happens in this function fairly well too. No need to comment for each line! (despite what uni tells you)

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,100 @@
...
+            msg = self._to_bytes_literal(data)
+
+            # Publish the data on the given topic
+            self.client.publish(topic, msg)
+
+            # Check for any pending messages on any of the subscribed topics
+            self.check_for_message()
+
+            # print("MQTT data sent: %s on %s through %s" % (msg, topic, self.mqtt_broker))
```
khanguslee: Don't really like commented out code. Either keep it or remove it. In this case, it might be good to have it?

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,100 @@
...
+
+        except OSError:
+            print('OSError, check line 82 of MQTT_Client_class.py')
+            # restart_and_reconnect()
```
khanguslee: Same here. Remove it or keep it.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def check_for_message(self):
+        """
+        Checks whether an incoming message from the MQTT broker on the subscribed topics is pending. This check is done
+        within the mqtt_pub method as well, but it gives a way to check manually for any
+        message (especially given that the MQTT Client is only held by this class).
+        :return: None
+        """
+        # If there's a message waiting, it will be sent to the callback_func defined in the connect_and_subscribe method
```
khanguslee: Probably don't need this comment as the docstring explains everything already.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,100 @@
...
+        self.client.check_msg()
+
+    def wait_for_message(self):
+        """
+        Waits for a message from one of teh subscribed topics. Note: This function will only finish executing when
+        a message is received and dealt in the set callback_func.
+        :return: None
+        """
+        # Keep checking for any incoming message/s until one is read
```
khanguslee: Same here.

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
+# Data Acquisition System - Middle wireless module
```
khanguslee: `Data Acquisition System - Middle Wireless Module`

- Capital letters for headers.

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+[![All Contributors](https://img.shields.io/badge/all_contributors-6-orange.svg?style=flat-square)](#contributors)
+
+The Middle Wireless Module allows us to collect and send data from x sensors attached to the bike. The middle wireless module consists of a micro controller (ESP32 currently) which gathers the data from the on board sensors and sends them using MQTT (using the internet).
```
khanguslee: Maybe put quotes here: `'x' sensors`

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+This repository contains all the code related to Monash Human Power's middle wireless module - part of their Data Acquisition System (DAS).
+
+## Getting Started
+
+1. Press clone on the root folder and use it to clone the entire repository on the command line.
+2. Create a config.py file in the same directory as the contents of this folder (refer below).
```
khanguslee: Can you use monospacing for filenames? `config.py`

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+3. The config.py file should hold the following variables (ensure the name matches):
+- ESSID : Should be assigned to a string containing the SSID of the network to connect to
+- PASSWORD: Should be assigned to a string containing password of the network
+- MQTT_BROKER: A string containing either the web server name or IP address which acts like the Mqtt broker
+- builtin_LED = PIN(2, PIN.OUT)
```
khanguslee: Code should be monospaced too: `builtin_LED = PIN(2, PIN.OUT)`

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+The Middle Wireless Module allows us to collect and send data from x sensors attached to the bike. The middle wireless module consists of a micro controller (ESP32 currently) which gathers the data from the on board sensors and sends them using MQTT (using the internet).
+
+This repository contains all the code related to Monash Human Power's middle wireless module - part of their Data Acquisition System (DAS).
+
+## Getting Started
+
+1. Press clone on the root folder and use it to clone the entire repository on the command line.
+2. Create a config.py file in the same directory as the contents of this folder (refer below).
+3. The config.py file should hold the following variables (ensure the name matches):
+- ESSID : Should be assigned to a string containing the SSID of the network to connect to
```
khanguslee: Monospace for code variables like this:

- `ESSID` : Should be assigned to a string containing the SSID of the network to connect to

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+- PASSWORD: Should be assigned to a string containing password of the network
+- MQTT_BROKER: A string containing either the web server name or IP address which acts like the Mqtt broker
+- builtin_LED = PIN(2, PIN.OUT)
+
+## Contents
+- boot.py: Contains function to connect the micro controller to the internet
```
khanguslee: Monospace for all filenames, topic strings and code etc.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,58 @@
+# Author: Kunj
+# Last updated: 1/08/2020
```
khanguslee: Don't need these two lines. We already use github to version control so we can already see this data. Would also be annoying to change every time.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,58 @@
...
+
+# Import relevant libraries
+import esp
+import network # To connect to the WIFI
```
khanguslee: Don't need explanation as to why you need these libraries.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,58 @@
...
+
+def do_connect(essid, password):
+    """
+    Tries to connect to the WIFI 5 times at max
+    :param essid: The WIFI name to connect to, ensure it's not in double quotes but single ones
+    :param password: The WIFI password to connect, ensure it's not in double quotes but single quotes
+    :return: None
+    """
+
+    # How many attempts to connect to the WIFI
```
khanguslee: Watch out for being too verbose in your comments! Try to make sure the code that you write is as readable as possible. If this is not possible, then write a comment.

If you really wanted to make it clearer, you could change the variable to `max_wifi_attempts = 5` and this clearly tells the developer reading the code what this variable is, without needing to write an additional comment. Hope you get the idea!

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,91 @@
+# Note: You can remove all the PRINT statements when everything is already tested
+
+import machine
+# from wireless_module import WirelessModule
```
khanguslee: Why are these commented out?

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,91 @@
+# Note: You can remove all the PRINT statements when everything is already tested
```
khanguslee: Probably don't need this warning?

khanguslee (changes requested):
Congrats on your first PR 🥳 !!

Thanks for adding this in. Any documentation added to a codebase is always important. Although, there is a balance between having TOO much documentation in your code ;)

Can you also edit your PR title to be in imperative mood and explains what you've changed overall? This is what will be shown on the `git log` and `git blame`. See below: 
Screen Shot 2020-08-04 at 8 31 19 pm
We can see that this line was added when blake added sample micropython code. In this PR, bit hard to tell what `Kdav108/mid wireless module` means.

Your commit messages are so close to being in imperative mood btw! Instead of `Added functionality to connect to WIFI` change to: `Add functionality to connect to WIFI`. As simple as that!

Looking forward to many more PRs from you!

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: Yeah pretty sure it is a convention for docstring

wireless_modules/middle_module/boot.py
kdav108: yep that's a great idea, not sure why I didn't think of it

wireless_modules/middle_module/main.py
kdav108: I was planning to use them later, I guess I should have removed it till I start using them again

wireless_modules/middle_module/main.py
kdav108: I usually keep notes at the top of the code reminding what was only part of testing or temporary development use and can bee removed later. this way I can just control-F and delete them at the end. Not exactly a warning but in case we don't want print statements clouding the output later on.

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: True, fair point. But commenting is something I do solely for myself - I like it when comments can just spell out in English what most complex code bits do so I can debug and understand it much better (but from your feedback I can see I did over comment in a lot of places lol). And I guess for large projects it would make the files much longer and make the code more cloudy, so I'll try to keep it down in the future!

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
+# Note: Can remove all the Print statements once all testing is successful
+
+# Use micro pip to install packages
+import upip
+upip.install('micropython-umqtt.simple2')
+try:
+    from umqtt.simple import MQTTClient
+except FileNotFoundError:
+    print('Error in importing library')
```
hallgchris: I don't think this try-except is particularly useful, as the code won't be able to run if we can't import the library. I think we could just import it as normal.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class Client:
+    def __init__(self, client_id, mqtt_broker):
```
hallgchris: Something like `broker_address` would be more descriptive.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+                            data.
+        """
+        self.client = MQTTClient(client_id, mqtt_broker)
+        self.mqtt_broker = mqtt_broker
+
+    def connect_and_subscribe(self, topics_to_subscribe, callback_func):
+        """
+        Connects to the MQTT broker and subscribes to topic in 'topics_to_subscribe'
+        :param topics_to_subscribe: An array of topics to subscribe to.
+                                    Eah element must be a string or byte literal (the latter is preferred)
```
hallgchris: Typo: `Each`

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        :return: True if a connection to the Broker is successfully established, otherwise False
+        """
+
+        self.client.set_callback(callback_func)
+        try:
+            # Connect to MQTT broker
+            self.client.connect()
+            print('Connected to %s MQTT broker' % (self.mqtt_broker))
+        except OSError:
+            return False
```
hallgchris: While I appreciate attention to error handling, it would likely be useful to not catch a serious error like this, as if it occurs, we really want to see the traceback to understand the error. By just catching, we don't know anything about the error's cause.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        Converts 'data' into a form MQTT can read
+        :param data: A string of data to convert to bytes literal
+        :return: The bytes literal version of the 'data'
+        """
+        str_data = str(data)
+        return str.encode(str_data)
+
+    def mqtt_pub(self, data, topic):
+        """
+        This functions takes care of all of the formatting and publishes 'data' on the given 'topic'. It also checks for
```
hallgchris: Typo: `function`

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        # Subscribe to each topic
+        for topic in topics_to_subscribe:
+            self.client.subscribe(topic)
+            print('Subscribed to %s topic' % (topic))
+
+        return True
+
+    def _to_bytes_literal(self, data):
+        """
+        Converts 'data' into a form MQTT can read
```
hallgchris: This applies everywhere we're you've included a variable name or a fragment of code - chuck some backticks `\`` around it rather than single quotes, and it'll appear monospace when you hover over the function like so:
image

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        try:
+            msg = self._to_bytes_literal(data)
+
+            self.client.publish(topic, msg)
+
+            self.check_for_message()
+
+        except OSError:
+            print('OSError, check line 73 of MQTT_Client_class.py')
```
hallgchris: Again, better to let the error be thrown - that way we won't even have to worry about noting the line number, the traceback will do that for us.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        return str.encode(str_data)
+
+    def mqtt_pub(self, data, topic):
+        """
+        This functions takes care of all of the formatting and publishes 'data' on the given 'topic'. It also checks for
+        any pending message, but this should not be relied upon solely - use the check_for_message() method externally
+        when an incoming message is to be checked for/read.
+        :param data: A string of data to be sent
+        :param topic: A string representing the topic to send 'data' to.
+        :return: Nothing
```
hallgchris: Just for conciseness, I think you can omit the `return` line if there is no return value.

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        Checks whether an incoming message from the MQTT broker on the subscribed topics is pending.
+        :return: None
+        """
+        # If there's a message waiting, it will be sent to the callback_func defined in the connect_and_subscribe method
+        self.client.check_msg()
+
+    def wait_for_message(self):
+        """
+        Waits for a message from one of the subscribed topics. Note: This function will only finish executing when
+        a message is received and dealt in the set callback_func.
```
hallgchris: More concise: "This function blocks until a message is received"

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
...
+        :return: The bytes literal version of the 'data'
+        """
+        str_data = str(data)
+        return str.encode(str_data)
+
+    def mqtt_pub(self, data, topic):
+        """
+        This functions takes care of all of the formatting and publishes 'data' on the given 'topic'. It also checks for
+        any pending message, but this should not be relied upon solely - use the check_for_message() method externally
+        when an incoming message is to be checked for/read.
```
hallgchris: If this feature is not to be relied on solely, should it happen at all? I think the process of using the class would be more clear if `check_for_message()` was only called explicitly.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
+# Author: Kunj
+# Last updated: 1/08/2020
+
+# Import relevant libraries
```
hallgchris: Don't need to state this - we wouldn't even bother importing a library if it weren't relevant!

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
...
+import esp
+import network
+import config
+import time
+
+
+def do_connect(essid, password):
```
hallgchris: My main question for this is why do we only try 5 times max? Is there any downside to trying indefinitely? If we stop trying, then we're forced to always have to turn on the DAS on the bike before the wireless module, otherwise by the time we turn on the DAS the module may have stopped trying to connect.

I actually reckon https://github.com/monash-human-power/data-acquisition-system/blob/master/wireless_modules/mqtt_test/boot.py pretty much does what we want. Are you able to refactor the code so both `mqtt_test` and this can use the same function? Then you could call the existing function, and then turn on the LED when it returns, as by the time the function returns you are assured that we've made a connection.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
...
+    # Create station interface
+    station_interface = network.WLAN(network.STA_IF)
+    station_interface.active(True)
+
+    if station_interface.isconnected():
+        print('network config:', station_interface.ifconfig())
+
+        # Turn on the built in LED when booted and connected to WiFi (blue light)
+        config.builtin_LED.on()
+        time.sleep(5)
```
hallgchris: This sleeping will lengthen the boot process by 5 seconds. To be honest, we could probably just leave the light on as a connection indicator. In fact, can we make it turn off only when we disconnect from the network, maybe by checking `station_interface.isconnected()` elsewhere in the code?

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
...
+        config.builtin_LED.off()
+    else:
+        print('connecting to network...')
+
+        while not station_interface.isconnected() and max_wifi_attempts > 0:
+            # Connect to WiFi using the credentials given
+            station_interface.connect(essid, password)
+
+            if not station_interface.isconnected():
+                print('Could not connect to WIFI, attempts:', max_wifi_attempts)
```
hallgchris: I think you meant to to access the `attempts` variable here, which by the way was never initialised.

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
...
+            else:
+                print('Connected, network config:', station_interface.ifconfig())
```
hallgchris: LED doesn't turn on if we connect after the first attempt

wireless_modules/middle_module/boot.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+esp.osdebug(None)
+# connect to WIFI
+do_connect(config.ESSID, config.PASSWORD)
+
+
+
+
+
```
hallgchris: Can we remove all these blank lines?

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+try:
+    import config
+except FileNotFoundError:
+    print('Error importing config.py, ensure a local version of config.py exists')
+
+# To get access to the Client class in the MQTT_Client_class file
+try:
+    from MQTT_Client_class import Client
+except:
+    print("Error importing MQTT_Client_class.py, ensure a local version of the file exists")
```
hallgchris: `MQTT_Client_class.py`should always exist unless intentionally deleted, nevertheless I don't think we need an explicit try-except for this case, better to just throw the error rather than keep running without the import (which will cause a crash anyway)

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+# Define module number and MQTT topics to publish to
+MODULE_NUM = '2'
+PUB_DATA_TOPIC = b'/v3/wireless-module/{}/data'.format(MODULE_NUM)
+PUB_TOPIC_LOW_BATTERY = b'/v3/wireless-module/{}/low-battery'.format(MODULE_NUM)
+PUB_BATTERY_TOPIC = b'/v3/wireless-module/{}/battery'.format(MODULE_NUM)
+
+# Topics to subscribe to
+SUB_START = b'/v3/wireless-module/{}/start'.format(MODULE_NUM)
+SUB_STOP = b'/v3/wireless-module/{}/stop'.format(MODULE_NUM)
+SUB_TOPICS = [SUB_START, SUB_STOP]
```
hallgchris: Ideally we'd get topics from https://github.com/monash-human-power/common/blob/master/mhp/topics/topics.py, but we'll leave that for later as we need to work out how to use upip with a git python package.

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+    :return: None
+    :changes made: The global variable 'start' is changed to True or False depending on which topic we receive the
+                    message from
+    """
+    print('Successfully received message: ', msg, 'on:', topic)
+    if topic == SUB_START:
+        global start_publish
+        start_publish = True
+    elif topic == SUB_STOP:
+        global start_publish
```
hallgchris: Rather than having `global start_publish` twice, we can just have it once at the start of the function.

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+if not my_client.connect_and_subscribe(SUB_TOPICS, sub_cb):
+    # If the connect_and_subscribe returns False, a connection to the broker could not be established
+    restart_and_reconnect()
+
+# Wait for the start message
+my_client.wait_for_message()
+
+# Keep publishing data till a stop message is received
+test_no = 0
+while start_publish:
```
hallgchris: When we receive `stop`, we don't want to exit the program, instead we should keep listening for `start` again.

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # Keep checking for an incoming messages (specifically the stop message)
+    my_client.check_for_message()
+
+    utime.sleep(1)
+    test_no = test_no + 1
+
+print('########## Reached end of file successfully #########')
+
+
```
hallgchris, line 87: Rather than having everything here in the global scope, can we wrap it all up nicely into a class? The class could either inherit from your MQTT Client class, or just hold an instance as a member variable.

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+    description='MQTT wireless module test script with dummy payloads',
+    add_help=True)
+parser.add_argument(
+    '--module', action='store', type=int,
+    default=2, help="""The module number to determine the publish and subscription topics. 
+    If nothing is selected it will send on module number 2 topics.""")
+parser.add_argument(
+    '--host', action='store', type=str, default="5.196.95.208",
+    help="""Address of the MQTT broker. If nothing is selected it will
+    default to 5.196.95.208 - a test server by mosquito.""")
```
hallgchris: Can we default to localhost instead? I think it's much more likely that we would use a local broker. (Are you using a local broker? Let me know if you're not sure what this is)

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+    client.on_message = on_connect
+    client.subscribe(SUB_DATA_TOPIC)
+
+    # The start loop starts scanning for incoming messages and the stop loop stops this process.
+    client.loop_start()
+    for i in range(1):
+        client.subscribe(SUB_DATA_TOPIC)
+        # Wait more than 10 seconds in case, main.py could not connect to the broker and takes 10 seconds to restart
+        time.sleep(11)
+        # Send repeatedly to make sure it's received
```
hallgchris: Don't send things repeatedly for reliability - this is what [QoS](https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/) is for

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+
+    client.connect(BROKER)
+
+    # Calls the function on_connect whenever a message is received
+    client.on_message = on_connect
+    client.subscribe(SUB_DATA_TOPIC)
+
+    # The start loop starts scanning for incoming messages and the stop loop stops this process.
+    client.loop_start()
+    for i in range(1):
```
hallgchris: This loop will only run once

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+        client.subscribe(SUB_DATA_TOPIC)
+        # Wait more than 10 seconds in case, main.py could not connect to the broker and takes 10 seconds to restart
+        time.sleep(11)
```
hallgchris: Rather than forcing the user to wait 11 seconds, I think its fair to expect the user to run the script when they know the module is connected to the network. It would become annoying if you wanted to run the script a second time to have to wait another 11 seconds.

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+        # Send repeatedly to make sure it's received
+        print('Sending first start message')
+        for j in range(4):
+            client.publish(START_TOPIC, b'You can start now')
```
hallgchris: Our module doesn't understand English - just send an empty payload for clarity (you can omit the second argument entirely)

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+        # Wait for data to receive
+        time.sleep(10)
+
+        client.publish(STOP_TOPIC, b'You can stop now')
+        client.publish(STOP_TOPIC, b'You can stop now')
+        print('Sent stop message')
+    client.loop_stop()
+
+
+
```
hallgchris: Again, don't need all the newlines

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
...
+    default=2, help="""The module number to determine the publish and subscription topics. 
+    If nothing is selected it will send on module number 2 topics.""")
+parser.add_argument(
+    '--host', action='store', type=str, default="5.196.95.208",
+    help="""Address of the MQTT broker. If nothing is selected it will
+    default to 5.196.95.208 - a test server by mosquito.""")
+
+
+if __name__ == "__main__":
+    client = mqtt.Client('my_id')
```
hallgchris: If you leave out the client id, the library will choose a random client id, which is fine. By hardcoding the id we risk getting collisions if two people run the same script.

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,64 @@
+import paho.mqtt.client as mqtt
```
hallgchris: This file could be relevant to all modules, not just the middle one - consider moving it up a directory

wireless_modules/middle_module/MQTT_Client_class.py
```diff
@@ -0,0 +1,89 @@
+# Note: Can remove all the Print statements once all testing is successful
```
hallgchris: By convention we use lowercase filenames, and there's also no need to have the `_class` suffix. `mqtt_client.py` would be good!

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+[![All Contributors](https://img.shields.io/badge/all_contributors-6-orange.svg?style=flat-square)](#contributors)
+
+The Middle Wireless Module allows us to collect and send data from the sensors attached to the bike. The middle wireless module consists of a micro controller (ESP32 currently) which gathers the data from the on board sensors and sends them using MQTT (over the internet).
+
+This repository contains all the code related to Monash Human Power's middle wireless module - part of their Data Acquisition System (DAS).
+
+## Getting Started
+
+1. Press clone on the root folder and use it to clone the entire repository on the command line.
+2. Create a ``config.py`` file in the same directory as the contents of this folder (refer below).
```
hallgchris: Might have been a misunderstanding - you only need a single backtick on each side of the text you want to make monospace. `` `config.py` `` becomes `config.py`

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+[![All Contributors](https://img.shields.io/badge/all_contributors-6-orange.svg?style=flat-square)](#contributors)
+
+The Middle Wireless Module allows us to collect and send data from the sensors attached to the bike. The middle wireless module consists of a micro controller (ESP32 currently) which gathers the data from the on board sensors and sends them using MQTT (over the internet).
+
+This repository contains all the code related to Monash Human Power's middle wireless module - part of their Data Acquisition System (DAS).
+
+## Getting Started
+
+1. Press clone on the root folder and use it to clone the entire repository on the command line.
```
hallgchris: I would consider cloning the repo assumed knowledge, no need for this step! :)

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
+# DAS - MIDDLE WIRELESS MODULE
+
+[![All Contributors](https://img.shields.io/badge/all_contributors-6-orange.svg?style=flat-square)](#contributors)
```
hallgchris, line 3: We only need the all contributors stuff in the root readme, can remove from here (the stuff at the bottom of the file too)

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
+# DAS - MIDDLE WIRELESS MODULE
```
hallgchris: I Believe Angus Meant Tile Case (Like This)

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The Middle Wireless Module allows us to collect and send data from the sensors attached to the bike. The middle wireless module consists of a micro controller (ESP32 currently) which gathers the data from the on board sensors and sends them using MQTT (over the internet).
+
+This repository contains all the code related to Monash Human Power's middle wireless module - part of their Data Acquisition System (DAS).
+
+## Getting Started
+
+1. Press clone on the root folder and use it to clone the entire repository on the command line.
+2. Create a ``config.py`` file in the same directory as the contents of this folder (refer below).
+3. The ``config.py`` file should hold the following variables (ensure the name matches):
```
hallgchris: I would provide a `config.example.py` file like in `mqtt_test`, and instead put the documentation for each variable inside that file. Then, just provide instructions for duplicating and renaming that file here.

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+- ``ESSID`` : Should be assigned to a string containing the SSID of the network to connect to
+- ``PASSWORD``: Should be assigned to a string containing password of the network
+- ``MQTT_BROKER``: A string containing either the web server name or IP address which acts like the Mqtt broker
+- ``builtin_LED = PIN(2, PIN.OUT)``
+
+## Contents
+- ``boot.py``: Contains function to connect the micro controller to the internet
```
hallgchris: `functions` and `microcontroller

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+	- Also when running ``test.py`` if you do not receive any messages, re-run because most likely ``main.py`` was not ready to read the message from the start topic when it was sent by ``test.py`` (This may be due to some error and the need for the ESP32 to restart which would take some time)
+
+
+## Dependencies
+- The ``test.py`` requires the ``paho-mqtt`` library installed for Python
+- The other ``.py`` files are Micropython files so needs a board to run on (eg ESP32)
+- Ensure the ``config.py`` file is created locally (see the 'Getting Started' section above)
+
+
+## Steps to test?
```
hallgchris: `?` not required here

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+1. Download all the files (clone the root repository)
+2. Write a ``config.py`` file with the relevant variables (see above)
+3. Read this page (especially ensuring you understand what ``test.py`` does and what dependencies exist)
```
hallgchris: This duplicates the Getting Started section

wireless_modules/middle_module/README.md
```diff
@@ -0,0 +1,62 @@
...
+4. Run the ``boot.py`` file to get your ESP32 connected to the WIFI
+5. Run the ``main.py`` file and after starting it run the ``test.py`` file on your local computer and watch the console.
+6. Sample messages from the ``main.py`` file will be sent over to ``test.py`` which would output it to the console. The ``test.py`` file allows ``main.py`` to send for 10-14 seconds (default).
+7. Note that this process may take around 20 seconds (10 seconds for actual data sending and allowing for 10 seconds initially in case the ESP32 needs to reboot to reconnect to the MQTT broker)
```
hallgchris: This largely duplicates the Contents section

wireless_modules/middle_module/boot.py
hallgchris: In the rest of our python code we use `"`, so for consistency, that would be prefered

hallgchris (changes requested):
Unfortunately my ESP32 is still in the mail, so I can't test this :( but I'll assume it works!

There's some great work here! I've made some suggestions to clean up the code a bit so it's easier for people to use and follows our coding conventions. These code reviews will quickly become shorter as you get familiar with the sort of code we write!

Keep up the great work, let me know if you have any questions!

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: nice catch, even the IDE didn't catch it lol

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: I tried to dig into the error, but it seems it's just an error in connecting to the broker due to network instability, hence catching it and returning False allows the caller to call the method again, and in almost all of my cases it works the second time.

wireless_modules/middle_module/README.md
kdav108: agreed I'll remove it

wireless_modules/middle_module/README.md
kdav108: I tried to remove from these steps which I thought was overstating stuff but didn't really find much that's explicitly duplicated from the Contents section.

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: I'm a bit confused, I don't see any effect of having the backticks around the variable name when I hover over the function. Is this something that only works with viewing the code on GitHub maybe?

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: I'll remove that comment, it was only there to make sure no one uses the publish method to also do the checking for incoming message but I don't think anyone would do so.

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: This is more of a convention and appears automatically in most IDEs when you use doc strings. It kinda helps knowing straight away if you should be expecting a return value to deal with. If you feel it should still be taken out (especially because the method is quite small you can just check if it does or doesn't return anything) I can remove it, but then the doc string commenting will be inconsistent across methods.

wireless_modules/middle_module/boot.py
kdav108: Having infinite loops turned out to be the reason why I wasn't able to update my ESP32 using ampy or shell and I had to end up finding a way to remove the files that had the infinite loop. This error could have happened due to some other reason but from my knowledge it seems to be the infinite loop. So the 5 attempt (which was just arbitrarily chosen) was just a temporary fix. I was planning to look into getting the infinite loop working later, the priority at this point was to get the Mqtt class and test files out there so that others don't need to reinvent the wheel :)

wireless_modules/middle_module/main.py
kdav108: Yep that's true, but for the scope of this PR I thought I'll it leave it like this. Note that listening again also requires an infinite loop which didn't work for me so I still need to figure that out.

wireless_modules/middle_module/MQTT_Client_class.py
hallgchris: Yeah mine didn't either haha

wireless_modules/middle_module/MQTT_Client_class.py
hallgchris: Might be only in visual studio code. What editor do you use?

wireless_modules/middle_module/boot.py
hallgchris: That's very strange. Perhaps @Blake-Haydon had to deal with a similar issue?

wireless_modules/middle_module/main.py
hallgchris: No worries, since the mid-module stuff is still WIP. Sounds good!

wireless_modules/middle_module/test.py
kdav108: yep just left it there in case we need to do testing where we have to loop it more than once

wireless_modules/middle_module/test.py
kdav108: locahosts usually always work well, and since this was a test file I wanted to ensure that using an external host works as well. When u say we would use a local broker, do u mean in the end module or just when testing?

wireless_modules/middle_module/test.py
kdav108: Knowing when the module is connected to the network is kind of hard, a stable connection to the WIFI wouldn't guarantee a good connection to the MQTT broker, and it's in this latter case that the module is rebooted and takes some time. What we can do tho is remove that 11 sec gap and just re-run if we don't see any messages coming through?

wireless_modules/middle_module/main.py
kdav108: I was planning to design and add a class in the next PR when the main.py file actually has the function to read the sensor data and use it. But I can add a class now.

wireless_modules/middle_module/boot.py
kdav108: If we do this checking, I guess we don't need boot.py to infinitely try and connect to the WIFI, it can be done through other files that check that the interface is connected and if not calls on the do_connect() method

wireless_modules/middle_module/MQTT_Client_class.py
kdav108: I have checked it on VS code and PyCharm, just confirming in the screenshot, there's no backticks but the variable name is highlighted right?

wireless_modules/middle_module/boot.py
kdav108: On that note, I can create a class in here with two methods, one to connect (or attempt to) and another that can be called to verify the connection is still established, what do you think?

wireless_modules/middle_module/test.py
kdav108: Update: I tried using QoS 1 (QoS 2 is not supported by Micropython apparently) and in some instances it took me the third try to get the messages flowing through, this could be because the ESP32 wasn't checking for the message at the right time (but I assume QoS 1 would ensure that the packet is delivered so it would resend). So I would still suggest sending on the start topic multiple times (or sending again if we don't start receiving a message) just to make sure. And I don't really see any significant downside to doing so, unless I'm missing something here.

wireless_modules/middle_module/MQTT_Client_class.py
hallgchris: Oh right, in the code the variable is only highlighted I think because my cursor is inside it, this is just for the documentation that appears when you hover over the function name.

wireless_modules/middle_module/boot.py
hallgchris: Yeah that sounds like a good idea! Maybe put the class in a new file - importing the boot module from our main code feels a bit wrong to me

wireless_modules/middle_module/test.py
hallgchris: A local broker is when you host the MQTT broker on your own computer, using something like `mosquitto`. In the final product we will also be using a local broker, hosted on the DAS pi.

wireless_modules/middle_module/test.py
hallgchris: QoS 1 is fine, cos as you said it's fine to receive the message more than once. Were you both publishing with QoS 1 and subscribing with QoS 1? As far as I know, not that I've really used it much, MQTT will use the minimum qos of the publish/subscribe for when the message goes from the broker to the subscribed client.

I agree, there's no significant downside to simply sending the message multiple times, but if we can get it to work,
- Using QoS would be neater/more "correct"
- Using QoS guarantees us that the message will be delivered, but with sending multiple times there is a chance that none of those will arrive at all.

It's actually quite strange that you're having reliability issues at all. I do wonder if something's not set up right, or if there's something different about the ESP that I'm not aware of.

wireless_modules/middle_module/test.py
hallgchris: If we want to give the user this option, can you make it a command line argument? Generally we don't want the user to ever have to modify code in order to use something we make.

wireless_modules/middle_module/test.py
hallgchris: Yeah I reckon it's better that the user just reruns.

wireless_modules/middle_module/test.py
kdav108: Oh right, sure I'll make it local host then

wireless_modules/middle_module/test.py
kdav108: I'm using QoS 1 for both publishing and subcribing for the same reason you mention. 
When I talk about resending, we can resend with qos=1 (so I'm not suggesting remove using qos).
As for the set up and ESP32, I strongly believe the reason why it doesn't work sometimes is because the ESP32 sometimes fails to connect to the broker (I have seen this error appear a few times) and when this happens the ESP32 waits for 10 sec and reboots and then re-connnects and during this time the ESP32 is obviously not subscribed so even if we have qos=1 if the sender sends it before anyone subscribes, qos=1 doesn't help. 
However, this rebooting issues usually only occurs once and when you run it the second time it works, that's why I would suggest having a function which can check if the Mid module is sending any message and if not re-send the start message.

wireless_modules/middle_module/boot.py
hallgchris: The [micropython for esp32 documentation](https://docs.micropython.org/en/latest/esp32/quickref.html) even suggests using an infinite loop, so I feel like something else must have been causing your esp32 to lock up. Could you take another look at this?

wireless_modules/middle_module/MQTT_Client_class.py
hallgchris: I think we can still keep the documentation unambiguous if we know that no return type specified always means nothing is returned. If you really want to explicitly show nothing is returned, I highly recommend [type annotations](https://docs.python.org/3/library/typing.html) - so just add ` -> None:` to your function definitions.

wireless_modules/middle_module/test.py
hallgchris: I agree it sounds like a problem with the esp32 connecting to the broker. Rather than making a concession for this in the test code though, I think we really should focus on why the esp32 is failing to connect.

On the real bike, the `start` message will only be sent out once, and this test script needs to be able to ensure that the module is able to work consistently in this exact scenario. If the sensor module misses the start message, that's not the test script's problem, it has done it's job.

wireless_modules/middle_module/README.md
hallgchris: Looking back, I must have misread the readme. Ignore this!

wireless_modules/middle_module/boot.py
hallgchris: I actually reckon it would be good if we only had one `boot.py` shared by all the modules, since really all it has to do is connect to wifi using an example config, enable debugging, and turn on an LED when completed, which all modules will need to do. What are your thoughts on having `boot.py` up a directory and have both `middle_module` and `mqtt_test` share the one script?

wireless_modules/boot.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def do_connect(essid, password):
+    """
+    Continuously tries to connect to the WIFI
+    :param essid: The WIFI name to connect to, in string format
+    :param password: The WIFI password to connect, inn strung format
+    :return: Nothing
+    """
+
```
hallgchris: Maybe print something here to announce that you're starting to (attempt to) connect? This file looks much nicer now!

wireless_modules/middle_module/config.example.py
```diff
@@ -0,0 +1,9 @@
+from machine import Pin
+
+ESSID = "<WIFI NETWORK NAME HERE>"
+PASSWORD = "<WIFI PASSWORD HERE>"
+
+MQTT_BROKER = "<Boker domain address or IP address>"
```
hallgchris: Typo: `Broker`

wireless_modules/middle_module/main.py
```diff
@@ -3,25 +3,24 @@
...
-# To get access to the Client class in the MQTT_Client_class file
-from mqtt_client import Client
 
 # Define module number and MQTT topics to publish to
-MODULE_NUM = '2'
-PUB_DATA_TOPIC = b'/v3/wireless-module/{}/data'.format(MODULE_NUM)
-PUB_TOPIC_LOW_BATTERY = b'/v3/wireless-module/{}/low-battery'.format(MODULE_NUM)
-PUB_BATTERY_TOPIC = b'/v3/wireless-module/{}/battery'.format(MODULE_NUM)
+MODULE_NUM = "2"
+PUB_DATA_TOPIC = b"/v3/wireless-module/{}/data".format(MODULE_NUM)
```
hallgchris, line 15: Can you use f-strings here? e.g. `f"/v3/wireless-module/{MODULE_NUM}/data"`

We wouldn't be able to use byte strings with f strings, but I believe `paho-mqtt` is fine with passing normal strings for topics so it shouldn't be an issue.

wireless_modules/middle_module/main.py
```diff
@@ -51,35 +49,43 @@ def sub_cb(topic, msg):
...
+        client.mqtt_pub(message, topic)
+        print("MQTT data sent: %s on %s through %s" % (message, PUB_DATA_TOPIC, config.MQTT_BROKER))
+
+        client.check_for_message()
 
-if not my_client.connect_and_subscribe(SUB_TOPICS, sub_cb):
-    # If the connect_and_subscribe returns False, a connection to the broker could not be established
-    restart_and_reconnect()
+        utime.sleep(1)
+        test_no = test_no + 1
```
hallgchris, line 69: Stating this for other reviewers to see - this will only start/stop once before exiting which is a known issue, but since this PR is mainly for the MQTT client class and the rest is WIP we'll iron out stuff like this later.

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -20,44 +19,38 @@ def __init__(self, client_id, broker_address):
...
-        try:
-            # Connect to MQTT broker
-            self.client.connect()
-            print('Connected to %s MQTT broker' % (self.mqtt_broker))
-        except OSError:
-            return False
+
+        # Connect to MQTT broker
+        self.client.connect()
+        print("Connected to %s MQTT broker" % self.mqtt_broker)
```
hallgchris: Can you use f-strings in this function too?

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -20,44 +19,38 @@ def __init__(self, client_id, broker_address):
...
-        Converts ```data``` into a form MQTT can read
+        Converts data into a form MQTT can read
         :param data: A string of data to convert to bytes literal
-        :return: The bytes literal version of the ```data```
+        :return: The bytes literal version of the 'data'
         """
         str_data = str(data)
         return str.encode(str_data)
 
     def mqtt_pub(self, data, topic):
```
hallgchris: Just to neaten the API a little, could you put `topic` as the first parameter and then make `data` optional, defaulting to an empty string?

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -68,14 +61,11 @@ def mqtt_pub(self, data, topic):
...
-        :return: None
         """
-        # If there's a message waiting, it will be sent to the callback_func defined in the connect_and_subscribe method
         self.client.check_msg()
 
     def wait_for_message(self):
         """
-        This function blocks until a message is received
-        :return: None
+        This function blocks until a message is received on one of the subsrcibed topics
```
hallgchris: Typo: `subscribed`

wireless_modules/test.py
```diff
@@ -1,7 +1,7 @@
 import paho.mqtt.client as mqtt
 import time
 import argparse
-
+# Note: Change to local host at the end
```
hallgchris: Did you mean to remove this?

wireless_modules/test.py
```diff
@@ -1,7 +1,7 @@
...
 
 parser = argparse.ArgumentParser(
     description='MQTT wireless module test script with dummy payloads',
```
hallgchris, line 6: A few more single quotes in this file

wireless_modules/test.py
```diff
@@ -11,11 +11,16 @@
     default=2, help="""The module number to determine the publish and subscription topics. 
     If nothing is selected it will send on module number 2 topics.""")
 parser.add_argument(
-    '--host', action='store', type=str, default="5.196.95.208",
+    '--host', action='store', type=str, default="localhost",
     help="""Address of the MQTT broker. If nothing is selected it will
     default to 5.196.95.208 - a test server by mosquito.""")
 
 
+def on_connect(client, data, message):
```
hallgchris: Should this be called `on_message`? Also, I don't believe returning `True` means anything to `paho-mqtt`.

wireless_modules/test.py
```diff
@@ -41,21 +41,17 @@ def on_connect(client, data, message):
...
+    print('Sending start message')
+    # QOS=1 ensures the message is received at least once
+    client.publish(START_TOPIC, qos=1)
+
+    # Wait for data to receive
+    time.sleep(5)
+
+    client.publish(STOP_TOPIC, qos=1)
+    print('Sent stop message')
     client.loop_stop()
```
hallgchris, line 55: This is looking very nice now!

hallgchris (changes requested):
This is looking much nicer now! Thanks for your hard work looking into all the issues that were encountered. If you like, since this is branch is unlikely to get any major changes from now, you can branch of this branch and start working on the sensors.

@angustrau and @khanguslee this is much closer to completion now, worth another look. @Blake-Haydon would be great if you could try to run this on your ESP32, there's nothing hardware specific at this point.

wireless_modules/middle_module/main.py
kdav108: Sure, I'll check if normal strings are fine and if they are I'll use f-strings

wireless_modules/boot.py
```diff
@@ -0,0 +1,34 @@
...
+import esp
+import network
+import config
+
+
+def do_connect(essid, password):
+    """
+    Continuously tries to connect to the WIFI
+    :param essid: The WIFI name to connect to, in string format
+    :param password: The WIFI password to connect, inn strung format
```
khanguslee: nit: `in string`

wireless_modules/middle_module/main.py
khanguslee: I actually use this to highlight keywords such as `NOTE:`

https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight

wireless_modules/middle_module/main.py
khanguslee: One question I have is, how does the space consumed look like? Do we or are we near the storage capacity of the MQTT?

I remember when I was doing an Arduino Uno project, and I had too much code that I had to slim down to a string format that saves the most amount of space.

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,91 @@
...
+
+def restart_and_reconnect():
+    print("Failed to connect to MQTT broker. Reconnecting...")
+    utime.sleep(10)
+    machine.reset()
+
+
+def publish_test_messages(client, topic):
+    """
+    Keeps publishing data till a stop message is received
```
khanguslee: nit: `until` instead of till

wireless_modules/middle_module/main.py
khanguslee: perhaps add a `FIXME:` comment here?

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -0,0 +1,71 @@
...
+        Converts data into a form MQTT can read
+        :param data: A string of data to convert to bytes literal
+        :return: The bytes literal version of the 'data'
+        """
+        str_data = str(data)
+        return str.encode(str_data)
+
+    def mqtt_pub(self, data, topic):
+        """
+        This function takes care of all of the formatting and publishes 'data' on the given topic. Also checks for any incomming messages
```
khanguslee: spelling: `incoming`

wireless_modules/test.py
```diff
@@ -0,0 +1,57 @@
...
+    description='MQTT wireless module test script with dummy payloads',
+    add_help=True)
+parser.add_argument(
+    '--module', action='store', type=int,
+    default=2, help="""The module number to determine the publish and subscription topics. 
+    If nothing is selected it will send on module number 2 topics.""")
+parser.add_argument(
+    '--host', action='store', type=str, default="localhost",
+    help="""Address of the MQTT broker. If nothing is selected it will
+    default to 5.196.95.208 - a test server by mosquito.""")
```
khanguslee: No need to mention what it will default to. `argparse` will show you when you do a `--help` anyway. The address here is incorrect anyway.

khanguslee (changes requested):
Just a few spelling mistakes. But looks good otherwise!

wireless_modules/middle_module/main.py
kdav108: So it appears that Micropython actually doesn't support f-strings so we'll have to use .format()

khanguslee:
Side note: I like to read my files changes whenever I submit a PR to make sure everything is 100% btw. (For me it's mostly to remove `print` and `console.log` oops).

wireless_modules/middle_module/main.py
kdav108: Interesting, I'll have a look into it thanks!

wireless_modules/middle_module/main.py
kdav108: So I can make it continuously check for the start message again (The reason why it didn't work before was because of ampy (The thing that let me run scripts on ESP32) and so I have started using picocom instead which works with infinite loops). But I'll probably make the change in the next PR since this PR was mostly to get the MQTT class file out there

wireless_modules/middle_module/main.py
hallgchris: That's shame, thanks for checking though!

khanguslee (dismissed):
LGTM 💯

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -0,0 +1,72 @@
+
+# Use micro pip to install packages
+import upip
+upip.install("micropython-umqtt.simple2")
+from umqtt.simple import MQTTClient
```
hallgchris: We're not actually using the `umqtt.simple2` mqtt library which we install with `upip`, we're only using `umqtt.simple`. We can get rid of the `upip` stuff! (I'm aware this is an issue with `mqtt_test` as well, we can fix that another time)

hallgchris: Alternatively, if you want to do this in this PR (up to you), you could take a look at using `umqtt.robust2`. It has built in stuff for error handling and reconnecting when the broker is disconnected. Right now, if we loose connection to the broker we get a crash. Happy to do this later though, I imagine you want to get this PR out of the way! :)

wireless_modules/middle_module/mqtt_client.py
```diff
@@ -0,0 +1,72 @@
...
+    def _to_bytes_literal(self, data):
+        """
+        Converts data into a form MQTT can read
+        :param data: A string of data to convert to bytes literal
+        :return: The bytes literal version of the 'data'
+        """
+        str_data = str(data)
+        return str.encode(str_data)
+
+    def mqtt_pub(self, topic, data=""):
```
hallgchris: My bad, another thing for this function - Let's just call it `pub` or `publish`. `mqtt_` is redundant as we know this whole class is about mqtt! (I'd prefer `publish` because then it's consistent with both `umqtt.simple` and `paho-mqtt`)

hallgchris (changes requested):
I finally got my esp32! Played around a bit with it this morning and found out we're not even using the library that we're downloading. Other than that I'm happy with this PR!

wireless_modules/boot.py
```diff
@@ -0,0 +1,34 @@
...
+import config
+
+
+def do_connect(essid, password):
+    """
+    Continuously tries to connect to the WIFI
+    :param essid: The WIFI name to connect to, in string format
+    :param password: The WIFI password to connect, in string format
+    :return: Nothing
+    """
```
Blake-Haydon, line 14: These comments are much so much easier to read! Good work!

wireless_modules/boot.py
```diff
@@ -0,0 +1,34 @@
...
+    # Wait for the module to connect
+    while not station_interface.isconnected(): 
+        pass
+    
+    print("Connected, network config:", station_interface.ifconfig())
+
+
+# Allows the code to be uploaded with tools such as rshell / ampy
+esp.osdebug(None)
+do_connect(config.ESSID, config.PASSWORD)
```
Blake-Haydon, line 32: It would be really great if we could load all of the files in one go by using a shell script. There would be much nicer ways of doing this but something like this. (From the middle_module folder)
```
ampy -p $PORT -b 115200 put ../config.py
echo "main.py loaded"
ampy -p $PORT -b 115200 put main.py
echo "main.py loaded"
ampy -p $PORT -b 115200 put boot.py
echo "boot.py loaded"


# ... and so on for all the files
```

Blake-Haydon: I also think that if `boot.py` is in the global middle_module folder `config.py` should be as well as they can be used for all the modules.

wireless_modules/middle_module/config.example.py
```diff
@@ -0,0 +1,9 @@
+from machine import Pin
+
+ESSID = "<WIFI NETWORK NAME HERE>"
+PASSWORD = "<WIFI PASSWORD HERE>"
+
+MQTT_BROKER = "<Broker domain address or IP address>"
+
+builtin_LED = Pin(2, Pin.OUT)
+
```
Blake-Haydon: Looks good! Like before I would suggest moving this to the middle_module folder so it can be used for all the modules.

wireless_modules/middle_module/mqtt_client.py
Blake-Haydon: We still need `upip` when the board needs to download the micropython library for the first time

wireless_modules/middle_module/main.py
```diff
@@ -0,0 +1,89 @@
...
+try:
+    import config
+except FileNotFoundError:
+    print('Error importing config.py, ensure a local version of config.py exists')
+
+
+# Define module number and MQTT topics to publish to
+MODULE_NUM = "2"
+PUB_DATA_TOPIC = b"/v3/wireless-module/{}/data".format(MODULE_NUM)
+PUB_TOPIC_LOW_BATTERY = b"/v3/wireless-module/{}/low-battery".format(MODULE_NUM)
```
Blake-Haydon, line 16: Is there any chance we could change these to using the % format similar to `"Connected to %s" % self.mqtt_broker`. Just so that all of the code uses the same formatting style.

Blake-Haydon (changes requested):
Overall its looking really good, works exactly as intended in my testing! My only big grip is that it took me a while to get all of the code on the board. It would be fantastic if there was a shell script of make file to do all of the uploading at once so that it is easier to test.

wireless_modules/middle_module/config.example.py
hallgchris: I assume you meant the `wireless_modules` folder :joy: 
Now I think about it, `mqtt_client.py` can be moved too. The only thing specific to the middle module is `main.py`!

wireless_modules/boot.py
hallgchris: As a team, do you think we should generally use `rshell` or `ampy`? I though `rshell` was what we were using.

wireless_modules/middle_module/mqtt_client.py
hallgchris: That's strange. I was able to delete `import upip` and `upip.install(...)`, and delete all the downloaded libraries, and it still ran. Are you sure that `umqtt.simple` isn't installed by default?

kdav108: I did come across umqtt.robust2, and was planning to discuss if we should use it instead. Happy to look at it in this PR cos it directly affects the main purpose of this PR which is to get an MQTT Client class sorted. What do you think, worth going into in this PR?

wireless_modules/middle_module/mqtt_client.py
kdav108: yep agree

wireless_modules/boot.py
kdav108: Apparently ampy is much easier to use and most widely used from what I have seen around the web

wireless_modules/middle_module/config.example.py
kdav108: hahaha true, I agree with all this, it shall be moved :)

wireless_modules/middle_module/main.py
kdav108: Is there a preference for using % format, otherwise I think .format() is much easier to read

Blake-Haydon: I think `.format()` is easier as well. I just was the % format in other code blocks. Maybe we should change all to `.format()`.

wireless_modules/mqtt_client.py
```diff
@@ -0,0 +1,72 @@
...
+        self.client.set_callback(callback_func)
+
+        # Connect to MQTT broker
+        self.client.connect()
+        print("Connected to %s" % self.mqtt_broker)
+
+        # Subscribe to each topic
+        for topic in topics_to_subscribe:
+            self.client.subscribe(topic, qos=1)
+            print("Subscribed to %s topic" % (topic))
```
Blake-Haydon: Here is the `%` fomatting

wireless_modules/middle_module/mqtt_client.py
hallgchris: I reckon it's relevant so absolutely, go for it! I found I needed to use `import libs.umqtt.robust2`, that may or may not save you some time depending on if you've worked that out or not :joy:

kdav108: hahaha that probably did save a lot of my time, cheers. And in case if you have tried using robust2 already, did you have to still download umqtt.simple2, because apparently robust2 just builds on the simple2 class

hallgchris: I have a feeling that it downloads `simple2` automatically. No harm in trying without doing it manually, and adding it in if it doesn't work!

kdav108: > That's strange. I was able to delete `import upip` and `upip.install(...)`, and delete all the downloaded libraries, and it still ran. Are you sure that `umqtt.simple` isn't installed by default?

How did you delete the downloaded libraries from the board? All my attempts don't seem to work

Blake-Haydon (dismissed):
Amazing job!

wireless_modules/mqtt_client.py
```diff
@@ -70,3 +70,9 @@ def wait_for_message(self):
         This function blocks until a message is received on one of the subscribed topics
         """
         self.client.wait_msg()
+
+    def disconnect(self):
```
hallgchris, line 68: We don't seem to actually use this disconnect function anywhere. Did you mean to use it for something?

wireless_modules/mqtt_client.py
```diff
@@ -1,8 +1,8 @@
 
 # Use micro pip to install packages
 import upip
-upip.install("micropython-umqtt.simple2")
-from umqtt.simple import MQTTClient
+upip.install("micropython-umqtt.robust2")
+from umqtt.robust import MQTTClient
```
hallgchris: This isn't actually using `robust2` (https://github.com/fizista/micropython-umqtt.robust2), only `robust`! (https://github.com/micropython/micropython-lib/tree/master/umqtt.robust)

hallgchris (dismissed):
This still has some issues with regards to using `robust2` so that it doesn't crash when a broker is not found, but this PR is taking a long time and I imagine it's holding up your progress on work with getting sensor data. Since for the time being everything apart from this working I'm happy to let this through, the scope of this PR was probably getting away from us anyway. Will definitely need to revisit this though! We can't have the module crashing if the broker is unavailable.

Congrats on your first of many PRs though! :tada:

hallgchris:
@angustrau need your approval to merge. Feel free to disagree with what I've said :joy:

wireless_modules/config.example.py
```diff
@@ -0,0 +1,9 @@
+from machine import Pin
+
+ESSID = "<WIFI NETWORK NAME HERE>"
+PASSWORD = "<WIFI PASSWORD HERE>"
+
+MQTT_BROKER = "<Broker domain address or IP address>"
+
+builtin_LED = Pin(2, Pin.OUT)
```
emilytrau: Could initialising the pin be moved out of the config? So it's just the pin number in here

emilytrau (changes requested):

wireless_modules/config.example.py
kdav108: Yeah I can do that. Is this just so that config only has variables and no actual code?

emilytrau: Yep :+1:

wireless_modules/mqtt_client.py
kdav108: I realised MQTT allows you to disconnect from a broker as well. I don't see a use of it right now, but in case we need it in the future thought I'll put it in this class (especially since the main aim with this PR was just to get this class finalised).

wireless_modules/mqtt_client.py
kdav108: That's because there actually isn't anything called umqtt.robust2. If you look at the previous changes, even with umqtt.simple2, we only imported from umqtt.simple.

wireless_modules/mqtt_client.py
hallgchris: Yeah that makes sense, good thinking!

hallgchris (approved):
Did you mean to request my review for this? Cos this is looking good to me!

Blake-Haydon (approved):
LGTM 👍

emilytrau (approved):
LGTM

PR #59, opened 2020-08-09
Add MQTT radio bridge

hallgchris:
## Description

*begin readme copy-paste*

This program is used to join (or "bridge") the two MQTT brokers on the bike and the chase car so that MQTT messages may be sent without relying on being on the same local network.

At a high level, this works by making a subscription to all MQTT topics (`#`), and then for each received message:

- Check the message was not one that was sent by the bridge itself (performed in the `Bridge` class),
- Serialise the message into a series of bytes (performed in the `TxProtocol` class),
- Pack the bytes into one or more `Frame`s, or radio packets (also `TxProtocol`),
- Send each of the `Frame`s over the radio (using `ZetaRfRadio`).

On the other end of the bridge, the reverse process would happen.

- A packet is received by `ZetaRfRadio`,
- It is processed by `RxProtocol`, which will emit create a MQTT message when enough packets have arrived,
- `Bridge` will register the message as one sent by the MQTT radio bridge, before finally,
- The MQTT message is published using `MqttBridgeClient`.

Details on the protocol used for converting an MQTT message into bytes, and then into packets may be found on Notion.

*end readme copy-paste*

### Notes for reviewing

This is more of an MVP than a finished product. It works, but it has some notable flaws which must be fixed before we want to consider this remotely complete.
- If both ends of the bridge transmit a packet simultaneously, neither packet is successfully received. We'll likely need to add some sort of timing to our protocol to allow the ends of the bridge to take turns transmitting.
- There are no command line arguments or config files. The MQTT broker address is hardcoded to `tcp://localhost:1883` inside `main.cpp` if you wish to change it.
- Much of the code (the protocol stuff and the data structures) would be great candidates for unit tests. We should write some. Might need a third partly library? Need to research testing C++.
- I feel like there's something else I've forgotten.

I'm aware of the above problems and will save them for a future PR so we at least have an MVP merged for now. Other than that though, let me know of any and every suggestion/question you have!

If you're unfamiliar with C/C++ code: the `.hpp` files are "headers" and generally contain like an outline of the classes, with the names/signatures of their functions and their member variables. The `.cpp` files are where the functions are actually implemented and where all the logic happens. I'd start looking at the headers (which is also where most of the documentation is) to get an idea of what the class does, and then look at how it works in the `.cpp` afterwards.

Finally, I apologise for the size of the PR, I should have broken it up more, but I didn't realise the scope of the task when I started. I'll point out that the stuff in `Bridge/src/rf_configs` can be mostly ignored, especially the `.xml` file, as this is just configuration for the si4455 chip which is mostly auto-generated. (A bit more info on that in the readme.) Also, a lot of the code is documentation. I hope it makes sense and isn't too huge, let me know if anything doesn't makes sense or if you need more of a code tour.

## Screenshots

Receiving a hello world MQTT message in a debug build:

image

## Steps to Test

If you have a raspberry pi, you should be able to follow the steps in the readme.

If not, you may be able to at least compile and debug most of the code on your computer, excluding the Zeta radio section. I'm just guessing what should work so let me know if it doesn't.

- Install the Microsoft's CMake Tools vscode extension, and the C/C++ extension.
- Comment out (using `#`) the "Build/configure ZetaRF" section of `CMakeLists.txt`, as well as the line `"src/zeta.cpp"`
- Comment out (using `//`) lines 5 and 26 of `bridge.hpp`, and lines 10 and 21 of `bridge.cpp`. In other words, comment out all mention of the zeta radio.
- You should hopefully then be able to build/run/debug using the buttons on the vscode status bar.


Bridge/src/mqtt.cpp
```diff
@@ -0,0 +1,103 @@
...
+void MqttCallback::message_arrived(mqtt::const_message_ptr message)
+{
+    this->on_message_(message);
+}
+
+
+MqttBridgeClient::MqttBridgeClient()
+{
+    const auto client_id = this->generate_client_id();
+    this->client_ = std::make_shared<mqtt::async_client>("tcp://localhost:1883", client_id);
```
emilytrau: Is there a blocker on making this an argument?

README.md
```diff
@@ -8,15 +8,19 @@ This repository contains all the code related to Monash Human Power's Data Acqui
 
 ## Getting Started
 
-1. Type `git clone https://github.com/Monash-Human-Power/MHP_DAS.git` on the command line to download the whole repository.
+Type `git clone https://github.com/Monash-Human-Power/MHP_DAS.git --recurse submodules` on the command line to download the whole repository and submodules.
```
emilytrau: This should be `--recurse-submodules`

emilytrau (changes requested):
You've done a fantastic job on this Chris!

Bridge/src/mqtt.cpp
hallgchris: Good point, I could pretty easily make it an argument of `MqttBridgeClient`. I'll do that now!

Reason I didn't make it a command line argument is because afaik C++ doesn't have anything like argsparse built in. I imagine I'd get some 3rd party library to handle arguments.

hallgchris:
Thanks @angustrau!

Not to take all the credit, Angus did a fantastic job laying the groundwork for the radio protocol!

Bridge/src/mqtt.cpp
emilytrau: Since we only have 1 argument right now, just using `argv[1]` should suffice

Bridge/README.md
```diff
@@ -0,0 +1,61 @@
+# MQTT Radio Bridge
+
+
+## About
+
+This program is used to join (or "bridge") the two MQTT brokers on the bike and the chase car so that MQTT messages may be sent without relying on being on the same local network.
+
+At a high level, this works by making a subscription to all MQTT topics (`#`), and then for each received message:
+
+- Check the message was not one that was sent by the bridge itself (performed in the `Bridge` class),
```
khanguslee: Use:

```
1.
2.
3.
```

Since we're following steps here technically.

Bridge/README.md
```diff
@@ -0,0 +1,61 @@
...
+- Serialise the message into a series of bytes (performed in the `TxProtocol` class),
+- Pack the bytes into one or more `Frame`s, or radio packets (also `TxProtocol`),
+- Send each of the `Frame`s over the radio (using `ZetaRfRadio`).
+
+On the other end of the bridge, the reverse process would happen.
+
+- A packet is received by `ZetaRfRadio`,
```
khanguslee: Same here please. Use a numbered list.

Bridge/src/ring_buffer.hpp
```diff
@@ -0,0 +1,78 @@
...
+    *this->head_++ = item;
+    if (this->head_ == this->buffer_.end())
+    {
+        this->head_ = this->buffer_.begin();
+        this->full_ = true;
+    }
+}
+
+template <class T, size_t N>
+bool RingBuffer<T, N>::contains(T item) const
```
khanguslee, line 67: From what I can see, we never remove items from the ring buffer right? So that would mean, we will keep old messages in the ring buffer. Could we potentially remove the message hash from the buffer once the receiver has received it? This should make sure we have space inside the ring buffer and tat this method `contains` does not take unnecessary amount of time to search for things.

haotongwang:
> If both ends of the bridge transmit a packet simultaneously, neither packet is successfully received. We'll likely need to add some sort of timing to our protocol to allow the ends of the bridge to take turns transmitting.

@hallgchris Is there some sort of error message that indicates there's a collision (e.g. `CollisionException`) or is it corrupted data (sent bits != received bits)? If it's the latter, we will probably need to implement checksums on the frames/packets/message.

As for dealing with collisions, maybe we can use a modified version of [CSMA/CD](https://en.wikipedia.org/wiki/Carrier-sense_multiple_access_with_collision_detection) (if there is a collision, send a jam signal, and wait a random amount of time before retransmitting). This could be a bit complicated to implement so it may be simpler to have some sort of scheduling (e.g. alternate sending packets) if the reduction in effective bandwidth isn't a problem.

Great work by the way! Unfortunately, it's too complex for me to satisfactorily review 😅

Bridge/src/zeta.cpp
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(2));
+    }
+    debug << "ZetaRf worker thread exiting..." << std::endl;
+}
+
+void ZetaRfRadio::process_zeta_events()
+{
+    const auto event = this->zeta_.checkForEvent();
```
khanguslee, line 68: Should we add:

```cpp
if (!event) {
	return;
}
```

So we don't need `event &` inside the below if statements.

Bridge/src/zeta.cpp
```diff
@@ -0,0 +1,116 @@
...
+    if (event & ZetaRf::Event::DeviceBusy)
```
khanguslee, line 73: Would a switch case statement be better here instead? Are we only expecting these two events btw?

Bridge/src/zeta.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <functional>
+#include <thread>
+#include <vector>
+
+#include <ZetaRf.hpp>
+
+#include "frame.hpp"
+#include "thread_queue.hpp"
+#include "rf_config/si4455_revb1_bidir_fixed_crc_pre10_sync3MHP_pay8_433mhz_40kbps.hpp"
```
khanguslee, line 19: Why choose this rf_config and not others? Would a comment explaining this be good here and/or what you can replace this with?

Bridge/src/protocol.cpp
```diff
@@ -0,0 +1,162 @@
...
+    this->next_part_count_ = 0;
+}
+
+std::optional<mqtt::message_ptr> RxProtocol::receive_packet(const Frame packet)
+{
+    if (packet.frame_type != FrameType::Message)
+        // Not implemented
+        return { };
+
+    if (!this->next_frame_count_)
```
khanguslee, line 23: Comment above this to state what this group of if statements does would be good.

Bridge/src/protocol.cpp
```diff
@@ -0,0 +1,162 @@
...
+        this->reset();
+        this->next_frame_count_ = packet.frame_count + 1;
+        debug << "Frame skipped" << std::endl;
+    } else
+    {
+        // Normal case, increment for next frame
+        (*this->next_frame_count_)++;
+    }
+
+    if (packet.part_count == 0)
```
khanguslee, line 42: Another comment here to generalise what this group of if statement does would be good too.

khanguslee (changes requested):
Fantastic PR and work by @hallgchris and @angustrau. Certainly impressed by the protocol and the C++ code here.

My C++ code isn't that great, but I do have a few questions in my review.

khanguslee:
@kdav108 this PR might be of interest to you because this is fairly low-level and related to hardware 💯

hallgchris:
> @hallgchris Is there some sort of error message that indicates there's a collision (e.g. `CollisionException`) or is it corrupted data (sent bits != received bits)? If it's the latter, we will probably need to implement checksums on the frames/packets/message.

The zeta module actually handles checksums (CRC) for us! It's quite likely that there's a status from the zeta module which says an error has occurred that I'm just ignoring, it could even be that the CRC is failing and that's how it knows something's up. But either way, due to the built-in CRC we never see any corrupted data, the packets are just discarded.

Also thank you for the collision handling suggestions! I was thinking of doing some sort of scheduling like your second suggestion, but I will read up on CSMA/CD too, and whatever else I can find. Thanks!

Bridge/src/ring_buffer.hpp
hallgchris: Discussed this a few days ago but for everyone else - We don't need to explicitly remove items from the ring buffer. We're only interested in storing the most recent N items, so when N items have been added the `RingBuffer` will start overwriting elements at the start of the array.

Note that `contains` will check all items between the start of the array, and either the end of the array (which is a constant size) or the up until the `head_` (which is always before the end of the array). So, `contains` always runs in O(N), N being the size of the array, independent of how many times we've called `put`.

Bridge/src/mqtt.cpp
hallgchris: Very good point! Done that now.

image

Edit: `BROKER_IP` is now `BROKER_ADDRESS`

Bridge/src/zeta.cpp
hallgchris: A switch won't work because `event` is an integer, not an enum, and it's possible to receive more than one events at once.

These are the only two events we're handling, but there's heaps more! https://github.com/GIPdA/ZetaRF/blob/v2/src/zetarf_radio.hpp

emilytrau (approved):
LGTM :+1:

Bridge/src/zeta.cpp
khanguslee: I still think we should need this even if we don't choose to have a switch case!

Bridge/src/zeta.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+void ZetaRfRadio::process_zeta_events()
+{
+    const auto event = this->zeta_.checkForEvent();
+
+    if (!event)
+        return;
+
+    if (event & ZetaRf::Event::DeviceBusy)
```
khanguslee, line 73: We can simply just do: `if (ZetaRf::Event::DeviceBusy)` here now because we check if `event` is falsy above!

Bridge/src/zeta.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        // DeviceBusy error usually means the radio module is unresponsive and
+        // needs a reset.
+        std::cout << "Error: Device Busy! Restarting..." << std::endl;
+
+        if (!this->zeta_.beginWithPacketLengthOf(PACKET_LENGTH))
+            std::cout << "ZetaRf begin failed after comm error." << std::endl;
+        this->zeta_.restartListeningSinglePacket();
+    }
+    if (event & ZetaRf::Event::PacketReceived)
```
khanguslee, line 83: Same here!

khanguslee (changes requested):
nearly there!

Bridge/src/zeta.cpp
hallgchris: I think you've missed what's happening here :wink: The events here use bit flags!

`ZetaRf::Event` is an enum type, of which `DeviceBusy` is a possible value. `DeviceBusy`, and the rest of the enums, all have a (compile-time constant) integer value, which are all powers of two (see [here](https://github.com/GIPdA/ZetaRF/blob/v2/src/zetarf_radio.hpp) again). So, all the values will, in binary, be all `0` except for a single `1`.

`event` is also an integer. If we look at its binary representation, it is able to represent any (and every) combination of ZetaRf events by simply summing all the event values we wish for it to contain. Lets say `event` was to contain `ZetaRf::Event::PacketReceived`, which has a value of `0b000 0100`. This would mean that `event` could be any integer, as long as the third last bit is a `1`, to indicate that `PacketReceived` event.

So, on these lines when we're checking for events, we actually take a bitewise and of `event` and the binary number which represents the event we're checking. So, `event & ZetaRf::Event::PacketReceived` will either be `0b0000 0100` if there's been a `PacketReceived` event, or `0b0000 0000` otherwise. The former is truthy and the latter is falsey, so we can use this as the condition for an if statement.

In summary, `event & ZetaRf::Event::PacketReceived` is how we actually check if an event has occurred, `ZetaRf::Event::DeviceBusy` on its own is just a constant enum/number. I realise now you probably know how this all works and just missed that it's a bitewise `&` and didn't know `Event` was an enum... I hope someone else finds this interesting though! It's called bit flags if you want to google.

khanguslee: Thanks for writing that all out! Definitely missed that it was a bitwise `and` and I was still stuck in the javascript heavens. But thanks for clearing that up, I always find bit operators quite interesting and a cool way of doing things.

khanguslee (approved):
LGTM! Worth a read for anyone who doesn't know much about coding with bit operators.

Bridge/src/zeta.cpp
hallgchris: Yeah they're super cool! Stores 32 booleans in the space of one integer!

hallgchris:
It took most of a year since its inception but ayy radio module MVP complete!

PR #60, opened 2020-08-16
MQTT logger V2

Blake-Haydon:
## Description
Adds robust MQTT logging and playback capabilities. This allows for all MQTT messages to be simulated at real time. Another step towards a fully virtual bike!

**This PR:**
✅ Logger is now general purpose 
✅ Playback _finally_ added  
✅ Proper exception handling

**Next PR:**
❌ Old logger needs to be removed
❌ Update ReadMe and fix file names `mqtt_playback_V2.py` -> `mqtt_playback.py`
❌ Add JSON un-wrapper to produce nice csv logs that play nice with excel

## Screenshots

`mqtt_playback_V2.py` verbose output -> (real time | topic | message) 
image

`mqtt_playback_V2.py` verbose output -> (real time | sped up time | topic | message) 
image

### Test the logger `mqtt_recorder_V2.py`
1. `cd  DAS`
2. `poetry shell`
3. Subscribe to sensor/# and test/# @ broker.hivemq.com for 10s
`python -m das.mqtt_recorder_V2 sensor/# test/#  -v --host broker.hivemq.com -v -t 10`
4. You should now see a log file in csv_data

### Test the playback `mqtt_playback_V2.py`
1. `cd  DAS`
2. `poetry shell`
3.  Playback 0001_log.csv at 30x speed using localhost MQTT broker
`python -m das.mqtt_playback_V2 das/csv_data/0001_log.csv -v -s 30`
4. You can check it works by running `mosquitto_sub -t "#"`


DAS/das/mqtt_recorder_V2.py
```diff
@@ -0,0 +1,56 @@
...
+    # Logger can run forever or for a specific time
+    if args.time != float('Inf'):
+        time.sleep(args.time)
+    else:
+        while True:
+            try:
+                pass
+            except KeyboardInterrupt:
+                # Graceful exit
+                print(f"\nData saved in {CSV_FILEPATH}")
```
hallgchris: Can we print the name of the logfile here as well?

hallgchris: I think it would make sense to print this inside `main_recorder.stop()`, so that
- We can make sure it only prints if writing the file was successful
- It prints also when we're running the logger for a fixed duration of time

DAS/das/mqtt_recorder_V2.py
```diff
@@ -0,0 +1,56 @@
...
+        verbose=args.verbose)
+
+    # Logger can run forever or for a specific time
+    if args.time != float('Inf'):
+        time.sleep(args.time)
+    else:
+        while True:
+            try:
+                pass
+            except KeyboardInterrupt:
```
hallgchris: Rather than only doing this when manually closed with a `KeyboardInterrupt`, we probably want to at least *try* to write out the data in the case of any exception. Should we use a try-finally thing here?

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+CsvConfig = {
+    "delimiter": ',',
+    "quotechar": '`',
+    "quoting": csv.QUOTE_ALL,
+    "skipinitialspace": True,
+    "fieldnames": ["time_delta", "mqtt_topic", "message"],
+}
+
+
+class Record:
```
hallgchris: I feel like `Recorder` (or even `Logger`?) would be a better class name, the verb `Record` feels like more of a function name.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+    topics : List(str)
+        A list containing the topic strings that are to be subscribed to
+    broker_address : str
+        The IP address that the MQTT broker lives on
+    verbose : bool
+        Specifies whether the incoming MQTT data and warnings are printed
+
+    Attributes
+    ----------
+    TOPICS: List(str)
```
hallgchris: What differentiates an attribute in capitals vs lowercase?

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # Create csv_folder_path folder if none exists
+        Path(csv_folder_path).mkdir(parents=True, exist_ok=True)
+
+        # Name the csv log file xxxx_log.csv where xxxx is a number
+        previous_log_num = 0
+        for filename in os.listdir(csv_folder_path):
+            try:
+                if int(filename[0:4]) > previous_log_num:
+                    previous_log_num = abs(int(filename[0:4]))
```
hallgchris: Why absolute value?

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+            except ValueError:
+                if self._VERBOSE:
+                    print(
+                        f"WARNING: {filename} should not be in {csv_folder_path}")
```
hallgchris: Maybe make a `log` function which checks `self._VERBOSE`, it's quite a common pattern in this file. You could make the function shared by both `Record` and `Playback` by making some superclass they both extend which contains the function.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        self._LOG_FILE_WRITER = csv.DictWriter(
+            self._LOG_FILE,
+            delimiter=CsvConfig["delimiter"],
+            quotechar=CsvConfig["quotechar"],
+            quoting=CsvConfig["quoting"],
+            fieldnames=CsvConfig["fieldnames"],
+        )
+
+        # Add headers for csv
+        self._LOG_FILE_WRITER.writeheader()
```
hallgchris: I reckon once we've been able to successfully write to the file (here), it would be good to say something like "Writing to logfile <filepath>", just because when running in non-verbose, there's very little feedback as far as whether things are working.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        """Play the logged data at a certain speed using an async function.
+
+        Parameters
+        ----------
+        speed : float
+            Speed multiplier to determine how fast to send out the data. A higher value means faster.
+        """
+
+        if self._VERBOSE:
+            print(f"⚡ Playback initiated at {speed}x speed ⚡")
```
hallgchris: I think we could print this even when not on verbose mode. Otherwise we don't see anything, and I think some form of feedback that it's working would be good!

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+            log_file,
+            delimiter=CsvConfig["delimiter"],
+            quotechar=CsvConfig["quotechar"],
+            quoting=CsvConfig["quoting"],
+            skipinitialspace=CsvConfig["skipinitialspace"])
+
+        self._log_data = []
+        for row in csv_reader:
+            # Convert time delta to float
+            row["time_delta"] = float(row["time_delta"])
```
hallgchris, line 231: Could you get these column headers from `CsvConfig`? If that's too hard to read, maybe you could define constants for the header titles, and use the constants both here and in `CsvConfig`.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+                    hostname=self._BROKER_ADDRESS)
+            except Exception as e:
+                print(f"ERROR: {e}")
+
+        publish_queue = []
+        for row in self._log_data:
+            publish_queue.append(_publish_aux(row))
+
+        # Load all async operation at the start using gather
+        await asyncio.gather(*publish_queue, return_exceptions=True)
```
hallgchris, line 292: This async stuff is really cool, but is there any reason we need it? I don't see how the way it's currently coded has any benefit over looking at each row iteratively, unless you feel that for some reason we need to support logs where the rows don't appear in chronological order.

I do appreciate the coolness though!

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        The current time used to produce time deltas
+    _LOG_FILE: `File`
+        Open log file object that can be written to
+    _LOG_FILE_WRITER: `csv.DictWriter`
+        Csv writter object that is used to write the data to the _LOG_FILE file
+    _client: `paho.mqtt.client`
+        MQTT client that connects to the broker and recives the messages
+    """
+
+    def __init__(self, csv_folder_path: str, topics: list = ["#"], broker_address: str = "localhost", verbose: bool = False) -> None:
```
hallgchris: Some of these lines are super long - even if we don't go for a hard 79 character limit, could you split these huge ones up a bit?

DAS/das/mqtt_playback_V2.py
```diff
@@ -0,0 +1,34 @@
...
+    description='MQTT playback',
+    add_help=True)
+
+parser.add_argument(
+    'filepath', action='store', type=str,
+    help="""Filepath of the csv log""")
+
+parser.add_argument(
+    '--host', action='store', type=str, default="localhost",
+    help="""Address of the MQTT broker. If nothing is selected it will
```
hallgchris: Nit, but "If no broker is specified" might make slightly more sense.

DAS/pyproject.toml
```diff
@@ -10,6 +10,8 @@ paho-mqtt = "^1.5.0"
 urllib3 = "^1.25.9"
 pandas = "^1.0.3"
 argparse = "^1.4.0"
+pathlib = "^1.0.1"
+asyncio = "^3.4.3"
```
hallgchris: These are part of standard python aren't they? Why do they need to be in `pyproject.toml`?

DAS/das/tests/test_logger.py
```diff
@@ -0,0 +1,36 @@
...
+    # Check that the cleanup worked
+    assert not os.path.exists(TEST_FOLDER)
+    assert not os.path.exists(os.path.join(TEST_FOLDER, "0001_log.csv"))
+
+
+def test_Playback_time_diff():
+
+    main_logger = logger.Record(TEST_FOLDER)
+    main_playback = logger.Playback(
+        os.path.join(DATA_FOLDER, "test_log_1.csv"))
```
hallgchris: You may have forgotten to upload this file. Maybe rather than having a random file sitting around in the repo, you could make this function generate the file?

DAS/das/tests/test_logger.py
```diff
@@ -0,0 +1,36 @@
+from das.utils import logger
```
hallgchris, line 1: We've got to figure these folders out... Despite this being in a `tests` folder, this is the only thing that's a unit test, the rest I guess are more like testing utilities?

hallgchris (changes requested):
This looks really great! Definitely agree it's neater than the previous logger.

I'm curious as to how you envision this being used on the bike. If it were always running in the background, everything would be in one log file. Is this what you intend? Cos I suppose the file could be chopped up in the JSON parsing stage you commented about. Otherwise we'd need to have something to start and terminate the `Record` class, which actually is entirely doable now I think about it.

I like your idea with the JSON parsing producing a new CSV after the fact. Makes the asynchronous nature of separate sensor modules easier to deal with. When this is done, we should make this occur automatically on the DAS so that we can still download useful logs from the das immediately after the ride so we can get right into seeing how fast we went!

DAS/das/mqtt_playback_V2.py
khanguslee: Hmm, that's weird. I always thought `argparse` would show you the default values when you type `--help`. Turns out you need to add the default arg formatter. seen here: https://stackoverflow.com/a/12151325/10947824

Can you implement that into your script and remove all strings that mention what the default is.

We'll need to do that in all scripts that we have in MHP later too..

DAS/das/tests/test_logger.py
```diff
@@ -0,0 +1,36 @@
...
+
+CURRENT_FILEPATH = os.path.dirname(__file__)
+
+# Sample logs are stored here to be used in tests
+DATA_FOLDER = os.path.join(CURRENT_FILEPATH, "test_csv_data")
+# Used to store the logs created by the tests
+TEST_FOLDER = os.path.join(CURRENT_FILEPATH, "csv_data")
+
+
+def test_Logger_make_csv_folder():
```
khanguslee: Great that you wrote unit tests! I don't think we've set up `pytest` in this repo yet. Can your next PR be setting up `pytest` so it runs your unit tests and also a Github Action to run them with every PR (see other repos for examples).

DAS/das/utils/logger.py
khanguslee: Think this line definitely is above the 79 characters per line limit LOL

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        # The logger object can subscribe to many topics (if none are selected then it will subscrive to all)
```
khanguslee: subscrive -> subscribe

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        self.TOPICS = topics
+        self._VERBOSE = verbose
+        self._START_TIME = time.monotonic()
+
+        # Create csv_folder_path folder if none exists
+        Path(csv_folder_path).mkdir(parents=True, exist_ok=True)
+
+        # Name the csv log file xxxx_log.csv where xxxx is a number
+        previous_log_num = 0
+        for filename in os.listdir(csv_folder_path):
```
khanguslee, line 97: Might be better to abstract this out into a private method so the constructor looks cleaner.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        else:
+            raise ConnectionError(
+                "Connection was unsuccessful, check that the broker IP is corrrect")
+
+        # Subscribe to all of the topics
+        try:
+            for topic in self.TOPICS:
+                self._client.subscribe(topic)
+                if self._VERBOSE:
+                    print(f"Subscribed to: {topic}")
```
khanguslee: Another possible future refactor for these scripts is to implement an actual logger so we don't have any `print()` statements and can properly display different levels of information.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+
+        if self._VERBOSE:
+            print(f"⚡ Playback initiated at {speed}x speed ⚡")
+
+        # Run the event loop to issue out all of the MQTT publishes
+        asyncio.run(self._publish(speed))
+
+    async def _publish(self, speed):
+        """Async function that collects all the necessary publish functions and gathers them to then be run by the
```
khanguslee, line 264: Be careful with the long lines.

khanguslee (changes requested):
Yup just to clarify, we now have all topics recorded into a single file. I am fine with this, it just means we need to create a new python script that splits our CSV file into individual wireless modules via the topics. I actually like this format a lot better than previously as it is much easier to replay logs.

We might need to format this repo soon so it conforms with the other repos.. Anyone want to volunteer?

DAS/das/tests/test_logger.py
khanguslee: Whatever isn't a test inside the `tests` folder should relocated into the utils folder for now I reckon. (different PR)

DAS/das/mqtt_recorder_V2.py
Blake-Haydon: Yes it does this correctly now!

Blake-Haydon: Moved into the .stop() method

DAS/das/utils/logger.py
Blake-Haydon: When you call the function it will be `logger.Record()`. It used to be `logger.Logger()` but I was getting confused when working on it. I'd be fine to change it to `Recorder` if others think its better

DAS/das/utils/logger.py
Blake-Haydon: Capitals were meant to be const. Do you reckon it just makes it more confusing?

DAS/das/utils/logger.py
Blake-Haydon: I was thinking it could potentially be a negative filename which would mess it up with the hardcoding. It has been changed to a regular expression.

DAS/das/utils/logger.py
Blake-Haydon: I originally implemented it without async using `time.sleep`. I expected there would be a small difference between playback and actual time. Turns out there was a difference of 1s every minute. So for an hour recording we would be about 1m behind if we played it back.

The implementation in async is also much easier to understand and requires less math.

DAS/das/tests/test_logger.py
Blake-Haydon: Really good idea!

DAS/das/tests/test_logger.py
Blake-Haydon: Yeah definitely need to move them into utils

DAS/das/utils/logger.py
Blake-Haydon: haha, just a few over. Is there any good code formatters for python. Is black any good?

DAS/das/utils/logger.py
Blake-Haydon: Just added the real python logger.

I looked into using it to do the real logging functionality, but its not designed to record data. It is only good at logging human readable output.

DAS/das/utils/logger.py
hallgchris: `Record` isn't really a function though, it's a class. This becomes more obvious if you import it differently, so that it's not prefixed with the module name:
```python
from das.utils.logger import Record()
main_recorder = Record()
```
Considering the variable you're assigning the class to also contains `recorder`, I do kinda feel like the class should be called that and the assignment would make more sense. Happy to hear more opinions though!

DAS/das/utils/logger.py
hallgchris: Yeah that makes sense! Just hadn't really seen it in member variables before but I see nothing wrong with it. Should `_client` also be in capitals then?

DAS/das/utils/logger.py
hallgchris: It could be implemented fairly easily with `time.sleep`, just by storing the time which the last message was sent and then calculating how long until the time the next message should be sent. I think my main concern with async is that spawning thousands of threads feels messy and could effect performance (maybe, unsure) and feels unnecessary when the same logic could be achieved with a loop that people newer to coding would understand.

DAS/das/utils/logger.py
hallgchris: Yeah black seems to work well!

Blake-Haydon:
Almost all of the recommendations have been implemented. Standalone tests have also been written.

DAS/das/mqtt_recorder_V2.py
hallgchris: Any thoughts on this?

DAS/das/tests/test_logger.py
```diff
@@ -1,36 +1,149 @@
 from das.utils import logger
 import shutil
 import os
+import time
+import unittest
```
hallgchris, line 5: Any reason for going with `unittest` over `pytest`? Not totally against it! But asking because it might be better to be consistent with the rest of our repos which use `pytest`.

hallgchris: Also how do you run the tests? :sweat_smile: 

Would be good to document in readme

hallgchris (changes requested):
This is looking really good! Great unit tests and I like the `logging` library, maybe we need to use it elsewhere too. Not much more to add other than the stuff above!

DAS/das/tests/test_logger.py
Blake-Haydon: You run tests by just writing `pytest`. `pytest` can be used as just a runner for ordinary `unittest` tests. I used `unittest` as it requires building up a test and then tearing it down and its much easier to use `unittest`.

hallgchris: Oh right, did not know you could do that!

hallgchris (dismissed):
Sorry this took a while! Assuming you've made sure you've looked through my prior replies and don't agree with them (which is fine, maybe drop a reply as to why) then happy to approve :)

DAS/das/utils/logger.py
khanguslee: Yeah i'm happy for it to be used as a way to log our debug statements for now!

DAS/das/utils/logger.py
khanguslee: Had a good think about this. After reading the implementation inside `mqtt_recorder_v2.py` I think it would be good to use it like:

```python
    main_recorder = logger.Recorder(
        CSV_FILEPATH, topics=args.topics, broker_address=args.host, verbose=args.verbose
    )
	main_recoder.start() # or main_recoder.play()
```

This is to match with the `main_recorder.stop()`. It's not quite intuitive to know when we start recording. This should also solve the problem of `Record` vs `Recorder` too.

DAS/das/mqtt_playback_V2.py
```diff
@@ -0,0 +1,47 @@
...
+    help="""Address of the MQTT broker.""",
+)
+
+parser.add_argument(
+    "-s",
+    "--speed",
+    action="store",
+    type=float,
+    default=1,
+    help="""Playback speed up. Default is 1x""",
```
khanguslee: Can remove `Default is 1x` since you added the default formatter in argparse:

Screen Shot 2020-09-05 at 4 01 37 pm

DAS/das/mqtt_recorder_V2.py
khanguslee: > Should we use a try-finally thing here?

Sounds like a good idea to me!

khanguslee (changes requested):
SO CLOSEEEE. Added a few comments to existing discussions.

DAS/das/mqtt_recorder_V2.py
Blake-Haydon: I added a graceful_exit function which if an error occurs it will gracefully exit saving the last data

DAS/das/utils/logger.py
Blake-Haydon: Totally agree, added a start function. It makes it significantly easier to work with!

Blake-Haydon:
I think everything is looking good now!

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # Connect to MQTT broker
+        self._CLIENT = mqtt.Client()
+        self._CLIENT.on_connect = self._on_connect
+        self._CLIENT.on_message = self._on_message
+        self._CLIENT.connect(broker_address)
+        self._CLIENT.loop_start()  # Threaded execution loop
+
+    def _create_log_file(self, csv_folder_path: str) -> None:
+        """Is used to open a log file and csv obj during the object init
```
khanguslee: nit: Imperative mood for any method description btw!

Example:
`Generates a log file ready to be written in.`

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,275 @@
...
+    def __init__(
+        self,
+        csv_folder_path: str,
+        topics: list = ["#"],
+        broker_address: str = "localhost",
+        verbose: bool = False,
+    ) -> None:
+        # The logger object can subscribe to many topics (if none are selected then it will subscribe to all)
+        self.TOPICS = topics
+        self._VERBOSE = verbose
```
khanguslee: Instead of having the `if self._VERBOSE:` everywhere, you can actually tell the `logger` library to hide certain levels of logging (so can automatically hide all the info and debug logs).

``` python
if (verbose):
    logging.getLogger().setLevel(logging.ERROR)
```

I think the above might work. pls test.

DAS/das/utils/logger.py
```diff
@@ -0,0 +1,275 @@
...
+            scaled_sleep = row["time_delta"] / speed
+            await asyncio.sleep(scaled_sleep)
+
+            if self._VERBOSE:
+                logging.info(
+                    f"{round(row['time_delta'], 5): <10} | {round(scaled_sleep, 5): <10} | {row['mqtt_topic']: <50} | {row['message']}"
+                )
+
+            try:
+                publish.single(
```
khanguslee: Kinda inclined to change the alias for `publish` to `mqtt_publish` or `publish_mqtt` (or something similar). I wasn't sure where this function came from since it's pretty similar to `_publish`.

khanguslee (changes requested):
I'm pretty happy overall. I've tested it and it works amazingly too!

Just a few things that you probably didn't know about logging, but otherwise it's on the right track!

DAS/das/mqtt_recorder_V2.py
```diff
@@ -52,22 +52,25 @@
...
+        if args.time != float("Inf"):
+            time.sleep(args.time)
+        else:
+            while True:
                 pass
-            except KeyboardInterrupt:
-                graceful_exit(main_recorder)
+
+    except (KeyboardInterrupt, Exception):
+        pass
```
hallgchris: This means if an exception is thrown we won't ever get to see what happened. Maybe something like
```python
    except KeyboardInterrupt:
        pass
    except Exception as e:
        print(f"{type(e)}: {e}")
```
I know we're not likely to have anything go wrong in that try section but I'm just a little scared by catching all exceptions and ignoring them.

hallgchris (changes requested):
Just one more thing to add on to what Angus said! We're nearly there

DAS/das/mqtt_recorder_V2.py
Blake-Haydon: 100% agree will implement!

DAS/das/utils/logger.py
Blake-Haydon: I've switched to using a client instead of just a single publish statement. This resolves this issue while also making the error messages much more clear!

DAS/das/utils/logger.py
Blake-Haydon: Works a treat!

Blake-Haydon:
Only 3 main updates from the previous version.
- Imperative mood for any method description (checked and all fixed)
- Removed `self._VERBOSE` everywhere (much cleaner)
- Added more exceptions to catch when it doesn't connect to MQTT broker

Blake-Haydon:
Oh and all the tests work when running pytest 🎉

hallgchris (approved):
LGTM! Thanks for this!

khanguslee (approved):
ayyy LGTM! Looks good now 💯

PR #61, opened 2020-09-04
Middle Wireless Module sensor readings

kdav108:
## Description

The PR adds classes to read from the MQ135 (co2 concentration), DHT22 (temperature and humidity values) and MPU6050 (accelerometer and gyroscope values) sensors attached to the middle wireless module. It also adds code to collect all the data from each of these sensors and send them over the internet using MQTT. The classes should also be reusable for the other wireless modules.

After this PR the only software component of the middle wireless module left will be reading the battery percentage level.

## Screenshots

This is the middle wireless module sending data and the `test.py` file receiving and displaying those data sets:
Screen Shot 2020-09-04 at 2 01 24 pm

## Steps to Test

1) Upload the files to your ESP32 by running the command `./upload.sh <port_name>"` and replacing the port name placeholder with the name of your ESP32 device.
2) Run the `main.py` file using either ampy (`ampy -p <port_name> -b 115200 run main.py`), rshell or picocom (`picocom /dev/tty.SLAB_USBtoUART -b 115200`)
3) Run the `test.py` file inside the middle_module folder or 
4) use Mosquitto from the terminal to publish to the start (`mosquitto_pub -h broker_name -t /v3/wireless-module/2/start -m ""`) and stop (`mosquitto_pub -h broker_name -t /v3/wireless-module/2/start -m ""`) message and subscribe to the data topic manually (`mosquitto_sub -h broker_name -t /v3/wireless-module/2/data`).

Note: It might be a bit hard to test the program without the sensors, but if you have any of the three sensors mentioned above you could connect it to your ESP32 and modify the `main.py` file to remove the other sensors and connect your one, it should be quite straightforward. Otherwise, you could use a publicly hosted MQTT server, such as "broker.hivemq.com" and let me know when you want to test the sensor values and I can power up the wireless module I have :) This way you won't necessarily get to play around with the sensor and check the values but at least get some sense of the workings.


wireless_modules/common_sensors/dht_sensor.py
```diff
@@ -0,0 +1,40 @@
...
+        Reads the temperature and humidity measurements from the sensor
+        :return: An array of length 2 containing dictionary for the temperature and humidity values. Each dictionary has
+        a `type` key associated with the name of the sensor (a string) and a `value` key associated with the value of
+        the sensor (an integer)
+        """
+        if self.queries_made == self.max_queries:
+            self.queries_made = 0
+
+        # read (fresh) data if this is the first query in the 2 second frame, otherwise use the previously read data
+        if self.queries_made == 0:
```
hallgchris: Could this be implemented using time rather than query counts? That way, if we ever decide to change how often `read` is called, the DHT22 is still polled every 2 seconds.

wireless_modules/common_sensors/dht_sensor.py
```diff
@@ -0,0 +1,40 @@
+import dht
+
+
+class DhtSensor:
```
hallgchris: Would be good to mention somewhere in this file why we only poll every 2 seconds! Perhaps link https://docs.micropython.org/en/latest/esp8266/tutorial/dht.html

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        self.pub_topics.append(data_topic)
+        self.pub_topics.append(battery_topic)
+        self.pub_topics.append(low_battery_topic)
+
+    def add_sensors(self, sensor_arr):
+        """
+        Store instances of sensor class
+        :param sensor_arr: An array of sensor class instances.
+        :pre-requisites: Each class must have a .read() method to read data through
```
hallgchris: To make this more explicit, I would create a `Sensor` abstract base class from which all sensors inherit. This class would have an abstract `read` method, enforcing that all sensors have this method available. Then, this method only requires that `sensor_arr` is an array of `Sensor`s.

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        self.sensors = sensor_arr
+
+    def _read_sensors(self):
+        """
+        Reads sensor data from each sensor object stored within this class instance
+        :return: A dictionary of all the sensor types and they're corresponding sensor reading/s
```
hallgchris: Typo: `their`

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+            self.mqtt.wait_for_message()
+
+            while self.start_publish:
+                data = self._read_sensors()
+                print("-------Publishing--------")
+                self.mqtt.publish(self.pub_topics[0], ujson.dumps(data))
+                print("MQTT data sent: {} on {}".format(data, self.pub_topics[0]))
+
+                self.mqtt.check_for_message()
+                utime.sleep(data_rate)
```
hallgchris: This will cause data to be published every (1 + time to read/publish data) seconds. Instead, the module should wait here until 1 second (or whatever `data_rate` is) has elapsed since the start of the loop iteration. (there might be other ways of implementing/writing this that do the same thing - do whatever you see as neatest)

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+    def __init__(self, client_id, broker):
+        """
+        Initialises the wireless module
+        :param client_id: A string representing the mqtt client id
+        :param broker: A string representing the IP address or web domain of the mqtt broker
+        """
+        self.sensors = None
+        self.start_publish = False
+        self.sub_topics = []
+        self.pub_topics = []
```
hallgchris: Rather than storing a list of topics, store each topic you need to keep track of separately and explicitly. This way, we always know what topic we're referring to. (compared to something like `self.pub_topics[0]`, where it's difficult to tell what this is for)

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class WirelessModule:
+    def __init__(self, client_id, broker):
+        """
+        Initialises the wireless module
+        :param client_id: A string representing the mqtt client id
+        :param broker: A string representing the IP address or web domain of the mqtt broker
+        """
+        self.sensors = None
```
hallgchris: Initialise this to an empty dictionary, so that its type is always valid. This way, we won't get a crash if `_read_sensors` is somehow called before sensors are added.

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
+"""
+A class structure to read and collate data from different sensors into a dictionary and send through MQTT
+"""
+import ujson
+import utime
+from mqtt_client import Client
+
+
+class WirelessModule:
+    def __init__(self, client_id, broker):
```
hallgchris: I reckon you could generate the topic strings in this class rather than in `main.py`. This has two benefits - you'll only need to pass in the module ID rather than 5 topics, and when we have more modules implemented, we won't need to duplicate the topic string generation. You could pass the module ID in here rather than in separate methods, ensuring that the topics are set when the class is created.

wireless_modules/common_sensors/dht_sensor.py
hallgchris: Actually, we do call `read` more than once a second, as it is called from the CO2 sensor!

wireless_modules/common_sensors/mq135_sensor.py
```diff
@@ -0,0 +1,85 @@
+from MQ135 import MQ135
+
+
+class co2:
```
hallgchris: Class should be upper case, `CO2`

wireless_modules/middle_module/main.py
```diff
@@ -1,89 +1,56 @@
...
-    publish_test_messages(my_client, PUB_DATA_TOPIC)
-
 
-run_module()
-print("########## Reached end of file successfully #########")
+# Generate a unique client_id used to set up MQTT Client
+CLIENT_ID = ubinascii.hexlify(machine.unique_id())
+
+# RZERO (for calibration of MQ135) found when the sensor was first 'activated'
+RZERO = 8.62
```
hallgchris, line 11: How did you get this value? Is it better than the library's default?

wireless_modules/common_sensors/mq135_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Reads the co2 concentration level. It uses the temperature and humidity data to get a more 'correct' reading, if
+        provided.
+        Assumes only co2 in the air.
+        :return: An array of length 1 containing a dictionary with the sensor type (string) and sensor value (int)
+        """
+        self._read_temp_humidity()
+
+        if self.temperature is not None:
+            ppm = self.mq135.get_corrected_ppm(self.temperature, self.humidity)
```
hallgchris: Maybe a silly question, but have you confirmed that blowing on the gas sensor increases the ppm readings?

wireless_modules/common_sensors/mq135_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        self.humidity = humidity
+
+    def _read_temp_humidity(self):
+        """
+        Read temperature and humidity data from the dht sensor if provided.
+        """
+        if self.dht_sensor_provided:
+            data = self.dht.read()
+            self.temperature = data[0]["value"]
+            self.humidity = data[1]["value"]
```
hallgchris: If you like, you could add a method which returns the temp/humidity directly, so you don't need to parse the dictionary. Might be a little cleaner!

wireless_modules/middle_module/README.md
```diff
@@ -20,7 +20,9 @@ This repository contains all the code related to Monash Human Power's middle wir
 ```
 
 ## Contents
-- `main.py`: Processes messages and sends sensor data through MQTT
+- `main.py`: Declares topics to publish and subscribe to
+- `mpu.py`: Class that helps read data from the MPU6050 sensor
+- `test.py`: A python script to test the sensor data received from the middle wireless module over MQTT
```
hallgchris, line 25: Would be good to mention under dependencies that you need to run `git submodule update --init` to install some of the sensor libraries.

wireless_modules/middle_module/main.py
```diff
@@ -1,89 +1,56 @@
...
+
+# Define all the Pin objects for each sensor
+scl_pin = machine.Pin(22)
+sda_pin = machine.Pin(21)
+dht_pin = machine.Pin(4)
+mq135_pin = machine.Pin(34)
+
+# Change ADC resolution to 10 bits consistent with the ESP8266 in order to use the MQ135 library
+adc = machine.ADC(mq135_pin)
+adc.width(machine.ADC.WIDTH_10BIT)
```
hallgchris: Could this be moved to the MQ135 sensor constructor?

wireless_modules/middle_module/main.py
```diff
@@ -1,89 +1,56 @@
...
+
+# Instantiate sensor objects
+my_mpu = Mpu(scl_pin, sda_pin, 100)
```
hallgchris: 100 samples seems like a lot, are that many really necessary?

wireless_modules/middle_module/mpu.py
```diff
@@ -0,0 +1,95 @@
...
+        accel_values = {
+            "x": all_data["AcX"],
+            "y": all_data["AcY"],
+            "z": all_data["AcZ"]
+        }
+        gyro_values = {
+            "x": all_data["GyX"],
+            "y": all_data["GyY"],
+            "z": all_data["GyZ"]
+        }
```
hallgchris: We should convert these into some normal units! @khanguslee what did we use on V2? Not sure if we should use Gs or m/s^2, and rpm vs deg/s vs rad/s.

For now, we can convert to Gs by dividing by 16384 and to deg/s by dividing by 131 [source](https://www.twobitarcade.net/article/3-axis-gyro-micropython/)

wireless_modules/middle_module/mpu.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return [{
+            "type": "accelerometer",
+            "value": accel_values
+            },
+            {
+            "type": "gyroscope",
+            "value": gyro_values
+            }]
```
hallgchris: Funky formatting here. Need to get some linting on this repo sometime!

wireless_modules/middle_module/test.py
```diff
@@ -0,0 +1,70 @@
+import paho.mqtt.client as mqtt
```
hallgchris, line 1: Does this replace the other `test.py` up a directory?

hallgchris (changes requested):
Awesome PR! Turns out I have an MPU-6050 lying around, hooked it up to the ESP32 and it's working flawlessly. Good job!

Some more general stuff about project structure - I reckon it would be neater to have all of the sensors in one folder, even if not all of them are used by multiple sensors. Just reckon it would be neater to sort the files by function rather than what module they're used in. Also, I reckon it would be good to put all the imported submodules under one directory, again just for organisation. Perhaps `libs` or `libraries`.

Again, nice work! Can't wait to see the code run on the module in person.

wireless_modules/common_sensors/dht_sensor.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, pin, queries=1):
+        """
+        Initialise the DHT sensor for temperature and humidity readings
+        :param pin: A Pin object
```
khanguslee: But what does this param represent apart from the type? Is there any way of representing types in this docstring style?

"The pin that this sensor is connected to"

wireless_modules/common_sensors/dht_sensor.py
```diff
@@ -0,0 +1,40 @@
...
+        :param queries: Number of times the sensor will be read every 2 seconds
```
khanguslee: https://github.com/monash-human-power/data-acquisition-system/pull/61/files#r483877180

To add on, a new dev will be confused as to why we chose 2 seconds if they read this sentence. Make sure your comments and code is always understandable by someone new!

wireless_modules/middle_module/mpu.py
```diff
@@ -0,0 +1,95 @@
...
+        :param samples: An integer representing number of readings to take the average of
+        """
+        i2c = I2C(scl=scl_pin, sda=sda_pin)
+        self.accelerometer = accel(i2c)
+        self.calibrated_values = []
+        self.samples = samples
+
+    def get_smoothed_values(self, n_samples=10, calibration=None):
+        """
+        code from: https://www.twobitarcade.net/article/3-axis-gyro-micropython/
```
khanguslee: Place this near the bottom of the docstring? Would rather see what the function does than credits (soz)

wireless_modules/middle_module/mpu.py
```diff
@@ -0,0 +1,95 @@
...
+        if calibration:
+            # Remove calibration adjustment.
+            for key in calibration.keys():
+                result[key] -= calibration[key]
+
+        return result
+
+    def calibrate(self, threshold=50, n_samples=10):
+        """
+        code from: https://www.twobitarcade.net/article/3-axis-gyro-micropython/
```
khanguslee: Put this at the bottom as well.

wireless_modules/middle_module/mpu.py
khanguslee: Looks like accelerometer is in Gs https://github.com/monash-human-power/data-acquisition-system/blob/898472c7c181059240082906f21c432f7df095d5/Teensy/DAS/SparkFunLSM6DS3.cpp#L358

Gyroscope is in deg/s: https://github.com/monash-human-power/data-acquisition-system/blob/898472c7c181059240082906f21c432f7df095d5/Teensy/DAS/SparkFunLSM6DS3.cpp#L350

wireless_modules/middle_module/upload.sh
```diff
@@ -7,3 +7,15 @@ ampy -p $1 -b 115200 put ../boot.py
 echo "boot.py loaded"
 ampy -p $1 -b 115200 put ../mqtt_client.py
 echo "mqtt_client.py loaded"
+ampy -p $1 -b 115200 put ../wireless_module.py
```
khanguslee, line 10: Don't know much bash scripts, but is there a way to store the file names for each file? Happy to leave as is for this PR, but worried this will get larger. Perhaps in a seperate PR?

wireless_modules/common_sensors/mq135_sensor.py
khanguslee: Can you also rename this file to `co2_sensor.py` instead?

wireless_modules/common_sensors/mq135_sensor.py
```diff
@@ -0,0 +1,85 @@
+from MQ135 import MQ135
+
+
+class co2:
+    def __init__(self, pin):
```
khanguslee: I think this class is a bit dependant on too many classes (`MQ135` and a `DHT` instance). What happens if we decide to use a completely new CO2 sensor? We would need to replace every instance of `mq135 ` inside this class which would be a hassle!

One way we could refactor this is to have a `CO2` class that takes in the mq135 co2 sensor class and then the mq135 co2 sensor takes in a temp/humidity sensor which are all a concrete implementation of the `Sensor` ABC suggested by @hallgchris here: https://github.com/monash-human-power/data-acquisition-system/pull/61/files#r483881011. That way, we can neatly swap between different CO2 and temp/humidity sensors.

```python
class DHTSensor(Sensor):
	def __init__(self):
		...

	def read(self):
		...
class Mq135Co2Sensor(Sensor):
	def  __init__(self, temp_sensor):
		...
	def read(self):
		...
class CO2(Sensor):

	def __init__(self, co2_sensor):
		...

	def read(self):
		...
```

LMK what you think. @hallgchris you too. (pls rename class names to whatever suits you)

khanguslee (changes requested):
Thanks for writing this up! We can definitely start to see the wireless module coming along nicely!

Just a few comments to minimise pain later on. LMK what you think.

wireless_modules/middle_module/mpu.py
hallgchris: Cool! I believe by default, the MPU-6050 has the acceleration range set to 2G, and gyro set to 250. I already messaged @MatthewMadsen earlier today asking what range would be required for measuring crash loads, but interesting to see we went for the highest possible for V2. Greater ranges apparently are less accurate, so we'd want to keep the range as low as is still useful.

Also, I have realised with a 1 Hz sample rate, we're probably not going to get much data from a crash... something to think about

wireless_modules/common_sensors/mq135_sensor.py
hallgchris: Sorry if I'm missing something here, but what functionality does the `CO2` class have here? Won't it just call `read` on `Mq135Co2Sensor`, making it redundant? (I'm sure that's not the case haha)

wireless_modules/wireless_module.py
kdav108: Not trying to defend the implementation currently, but is the process that relies on the data sent from the middle wireless module so sensitive to the data coming through in exactly one second?

wireless_modules/wireless_module.py
kdav108: Actually we can instead have an `exception` raised if `_read_sensors` is called without first setting the `sensors`. Since the `_read_sensors` is only called in the `run` method and once we enter that method there's no way of setting the sensors so no point running that method at all. Let me know what you think.

wireless_modules/wireless_module.py
kdav108: Sound good and in fact that was my initial thought, but I was thinking what if one of the module needs to publish or subscribe to different topic/s (perhaps in the future). Is this unlikely to happen?

wireless_modules/middle_module/main.py
kdav108: I put more information about this variable on our [notion page](https://www.notion.so/DAS-4126fc83762e45a1a3f5a5bdb2549bd6#34ef14ea17c44487b10fae53c8be5c2b) 
I'll summarise it here: `RZERO` differs depending on the sensor and can be found by reading the sensor for the first time after calibration. A guide for this can also be found [here](https://hackaday.io/project/3475-sniffing-trinket/log/12363-mq135-arduino-library):  (control F to jump to the section where RZERO is mentioned)

wireless_modules/common_sensors/mq135_sensor.py
kdav108: hahaha yeah I have, does the code look dodgy?

wireless_modules/wireless_module.py
hallgchris: I don't believe anything is actually sensitive to the data rate, I just reckon it would be nicer if the frequency that we publish is well defined. That way all sensors publish at the exact same rate, and we can say "data is published at 1Hz" rather than "every 1 and a bit seconds". It's nothing essential, but it's kinda like having clean code, it would just be a cleaner implementation to have this interval well defined :)

wireless_modules/wireless_module.py
hallgchris: I don't think there's any reason we should prevent the wireless module from running if no sensors are added. If we allow no sensors to be added, it lets people test the code on their ESP32 even if they don't have any sensors to connect to it (or they can't be bothered to connect them).

wireless_modules/wireless_module.py
hallgchris: It's unlikely to happen, but if it does, a neat solution would be to make a subclass of `WirelessModule` overriding whatever behavior is needed.

wireless_modules/middle_module/main.py
hallgchris: Thanks for the detailed info! :)

wireless_modules/common_sensors/mq135_sensor.py
hallgchris: Nah it looks alright! Was just interested :)

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,85 @@
+from MQ135 import MQ135
+
+
+class CO2:
+    def __init__(self, pin):
+        """
+        A MQ135 sensor class to read co2 concentration levels
```
nix-chu: Full stop here.

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        Provide the dht sensor to automatically read the temperature and humidity values to get a more accurate co2
+        concentration reading when the read() method is called.
+        :param dht_instance: An instance of the dht class (Must contain a .read() method)
+        """
+        self.dht = dht_instance
+        self.dht_sensor_provided = True
+
+    def set_temp_humidity(self, temp, humidity):
+        """
+        Sets the temperature and humidity measurement manually, used to get a more accurate co2 concentration reading.
```
nix-chu: Use imperative mood here so that the method acts like a command for the class. 

More info on docstring standards here: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        Read temperature and humidity data from the dht sensor if provided.
+        """
+        if self.dht_sensor_provided:
+            data = self.dht.read()
+            self.temperature = data[0]["value"]
+            self.humidity = data[1]["value"]
+
+    def get_rzero(self):
+        """
+        Reads the RZERO value from the MQ135 library
```
nix-chu: Needs imperative mood here.

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        self._read_temp_humidity()
+
+        if self.temperature is not None:
+            return self.mq135.get_corrected_rzero(self.temperature, self.humidity)
+        else:
+            return self.mq135.get_rzero()
+
+    def set_rzero(self, rzero):
+        """
+        Sets the RZERO value in the MQ135 library
```
nix-chu: Imperative mood and full stop here.

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+        :param rzero: An integer representing the Calibration resistance at atmospheric CO2 level
+        """
+        self.mq135.RZERO = rzero
+
+    def read(self):
+        """
+        Reads the co2 concentration level. It uses the temperature and humidity data to get a more 'correct' reading, if
+        provided.
+        Assumes only co2 in the air.
```
nix-chu: Both of these sentences need imperative mood.

wireless_modules/sensors/dht_sensor.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        self.sensor = dht.DHT22(pin)
+        self.readings = []
+
+        self.max_queries = queries
+        self.queries_made = 0
+
+    def read(self):
+        """
+        Reads the temperature and humidity measurements from the sensor
```
nix-chu: Imperative here.

wireless_modules/sensors/mpu.py
```diff
@@ -0,0 +1,95 @@
...
+        :param samples: An integer representing number of readings to take the average of
+        """
+        i2c = I2C(scl=scl_pin, sda=sda_pin)
+        self.accelerometer = accel(i2c)
+        self.calibrated_values = []
+        self.samples = samples
+
+    def get_smoothed_values(self, n_samples=10, calibration=None):
+        """
+        code from: https://www.twobitarcade.net/article/3-axis-gyro-micropython/
```
nix-chu: Capitalise here

wireless_modules/wireless_module.py
```diff
@@ -0,0 +1,100 @@
...
+        Reads sensor data from each sensor object stored within this class instance
+        :return: A dictionary of all the sensor types and they're corresponding sensor reading/s
+        :pre-requisite: The read() method for each sensor must return a dictionary
+        """
+        readings = {"sensors": []}
+
+        for sensor in self.sensors:
+            sensor_data = sensor.read()
+            for data in sensor_data:
+                readings["sensors"].append(data)
```
nix-chu, line 55: Forgive me if my understanding is incorrect, because I don't have an ESP32 to test this code.  But couldn't you just append `sensor_data` into the `"sensors"` array without needing to loop through `sensor_data`?

nix-chu (changes requested):
Just pointed out a lot about docstring errors here. There was a lot of times when you switch back and forth with an imperative mood and an indicative mood, which is super inconsistent and bad practice. I didn't comment on all of them since there was too many 😅 

Here is a useful document shared by Chris from a while back that explains the importance of using imperative mood (for Git commit messages, but also useful for method docstrings): https://chris.beams.io/posts/git-commit/#imperative

wireless_modules/wireless_module.py
hallgchris: This is done because a sensor may return multiple bits of data, e.g. the DHT22 returns both temperature and humidity. So, we need to loop through any returned data and add them separately.

wireless_modules/middle_module/mpu.py
kdav108: Do all sensors need to be queried at the same rate, otherwise we could query mpu6050 more often so we can gather better crash data. And if we don't want to affect other stuff by querying the mpu6050 more often, we could make the wireless module publish on a seperate topic for all mpu6050 data along with publishing on the existing data topic at 1Hz with all sensor values.

wireless_modules/middle_module/test.py
kdav108: We could, I thought of leaving the test.py up a directory for testing Mqtt topics but I just realised we may as well do that with the new test.py sine it's almost the same.

wireless_modules/middle_module/mpu.py
hallgchris: Yeah I reckon it's likely we'll want to do something like that in the future. For the scope of this PR though, let's just query it at the same rate as other sensors. Something for another PR would be increase the accelerometer and gyroscope's ranges, because I don't think the library supports that as is.

wireless_modules/sensors/co2_sensor.py
khanguslee: Just to add on, this is kind of a good practice to utilise. I've been in workplaces that use the following commit message style (https://www.conventionalcommits.org/en/v1.0.0/) which is a bit verbose for what we do, but it does further help with what each commit does.

Also, if everyone conforms to the same commit messaging style, I feel like the commit messages for the repo tend to be a lot cleaner and more succinct.

wireless_modules/common_sensors/mq135_sensor.py
khanguslee: > Sorry if I'm missing something here, but what functionality does the CO2 class have here? Won't it just call read on Mq135Co2Sensor, making it redundant? (I'm sure that's not the case haha)

@hallgchris yeah, you're right. I wanted a 'parent' class in which all CO2 sensors are under that way we can quickly swap the instance for another 'CO2' sensor. However, since it is a `Sensor` class already, it is a bit redundant.

wireless_modules/sensors/sensor_base.py
```diff
@@ -0,0 +1,11 @@
+"""
+Abstract base class for all sensors
+"""
+
+
+class Sensor:
```
khanguslee, line 4: Make this an actual ABC by doing:

```
from abc import ABC, abstractmethod

class Sensor(ABC):
    def __init__(self):
        pass

    @abstractmethod
    def read(self):
        pass
```

PS: Don't forget docstrings!

khanguslee (changes requested):
ayyyy only one extra comment that I have, but the code is looking good!

wireless_modules/sensors/sensor_base.py
hallgchris: Where's my syntax highlighting, I can't read this!

wireless_modules/common_sensors/dht_sensor.py
kdav108: In the current implementation it isn't called from the CO2 sensor. The CO2 sensor only provides the option to use the dht22 if it is provided.

kdav108: But regardless I'll try implement it with time instead

wireless_modules/sensors/sensor_base.py
kdav108: Woah I didn't know Python needed this 😅. And what doc string comment would be needed here? Like explain why the read method is an abstract method?

wireless_modules/middle_module/upload.sh
kdav108: Not an expert on bash scripts either lol but I do think I can save them in a variable. I can definitely look into it for a future PR!

wireless_modules/common_sensors/dht_sensor.py
kdav108: Hahah funny that in my last PR I had issues with over-commenting and in this one I'm missing key comments. 
Anyways, I added class comments just before the `_init_` so that should help explain it I hope

wireless_modules/common_sensors/mq135_sensor.py
kdav108: I would regardless have to parse it right? Either in an extra method in `dht_sensor` or in this class. I could add comments here or break the code down more so that it is more legible?

wireless_modules/sensors/sensor_base.py
nix-chu: If you're talking about docstrings for abstract _classes_, then I think it should just be the same as explaining what a class does. For me a good checklist would be:
- Say its an abstract class (so developers know it needs to be extended)
- Summarise what the class does or represents
- List important class/instance variables that are shared between its subclasses (in this case, none)
- How its meant to be implemented or how it should be used

But if you're talking about writing docstrings for an abstract _method_, then it should be fine to just write what it should do for its subclasses. Or don't and just write `pass`, that's fine too, because you can always write the docstring in the subclasses.

The following pics are from `data.py` in the Raspicam repo that I've been working on. I used these because the files were open for me at the time 😆 , but hopefully these can be good examples. 
image

image

nix-chu: Double-backed and saw what you currently have for `Sensor` and I think what you have currently works fine.

wireless_modules/common_sensors/mq135_sensor.py
kdav108: I feel like if we do use a different CO2 sensor in the future, we could just write another class for it and remove the current `CO2` class. The benefit of having a general parent `CO2` class, as I see it, is that if we use a different sensor in the future we can simply make changes here and won't have to change any other files, but in the current implementation even if we do make a new class for a new sensor we only need to change it in main.py (and I highly doubt that that the current `CO2` sensor will be used in any other files even in the future). LMK if I'm getting this wrong.

wireless_modules/sensors/sensor_base.py
kdav108: Thanks a lot for the detailed explanation @nix-chu! I was more confused because the sensor base class as it is didn't need much explaining with comments other than (in my opinion) explaining how the read method should be implemented. But I like the checklist you summarised above, it's quite handy!

wireless_modules/middle_module/main.py
kdav108: The guide I was following used 100 samples so I took that as a hint that 100 provides good results. However, I tried changing that to 50, 20 and 10 and they all seem to work well (at least on a relative basis). At the moment I left it at 20 (not too small and not too big) but maybe we should try out the number of samples that's ideal for us perhaps when we can actually get the module attached to a bike.

kdav108:
> ayyyy only one extra comment that I have, but the code is looking good!

So in regards to this comment about using the `ABC` library, just thought I'll let you guys know that Micropython doesn't support this library (Thanks to @hallgchris for pointing this out to me). There is a library in progress (not fully developed yet) and so me and Chris decided to use it just to convey the intent to anyone looking at this code later on. So just note that it won't actually enforce subclasses to have the read() method. 
Another note regrading this: I couldn't use upip to install this WIP library in the running code for some reason so I have manually downloaded it and included it in the `upload.sh` so it automatically places it on the board.

kdav108:
Or on second thought should I include the `abc.py` library as a submodule instead, so if it ever gets completed it would be easier to update it.

kdav108:
I have also updated `test.py` so that we can see the change in sensor readings for the MPU sensor. It just helps to see whether playing around with the module actually changes the accelerometer and gyroscope values.

wireless_modules/libraries/abc.py
```diff
@@ -0,0 +1,2 @@
+def abstractmethod(f):
+    return f
```
nix-chu, line 9: Could you perhaps write documentation as to why this exists and how to use it? Just say that MicroPython doesn't support the ABC module, so this file is needed to replace that for the moment. And also that it is used by the Python's decorator magic method '@'

wireless_modules/sensors/co2_sensor.py
```diff
@@ -0,0 +1,95 @@
...
+                ppm = self.mq135.get_corrected_ppm(self.temperature, self.humidity)
+
+                print("\t MQ135 Corrected PPM: " + str(ppm))
+
+            else:
+                ppm = self.mq135.get_ppm()
+
+                print("\t MQ135 PPM: " + str(ppm) + "\t rzero: " + str(self.mq135.get_rzero()))
+        except ValueError:
+            ppm = 0  # Since ppm cannot reasonably be 0, this can be used to mark that the sensor has not yet heated up
```
nix-chu: Could you put this comment above `ppm = 0`? Just reads easier from top to bottom. Also helps explaining why the try-except statement is here

nix-chu (changes requested):
Almost done! Just a few comments about the new things that you implemented

wireless_modules/libraries/abc.py
nix-chu: Also when using `from abc import abstractmethod`, would this user-created file conflict with the built-in Python module?

kdav108: Shouldn't conflict since there is no built-in module in Micropython for ABC. And the reason why I didn't comment the function was because I didn't write it (and I have heard you usually don't want to comment third-party code (since you want to leave it as it is so it can easily be compared and updated if needed). However, I maybe I'll link the source and perhaps describe why I'm using it at the start of the file.

nix-chu (approved):
LGTM!

wireless_modules/common_sensors/mq135_sensor.py
hallgchris: You're quite right, forgot that the data is stored as JSON internally. Perhaps we could store the raw data instead? Happy to go either way, as there's pros and cons of both methods

wireless_modules/middle_module/main.py
hallgchris: Sounds good! Will be good to see if there's any increase in noise with 20 samples.

wireless_modules/libraries/abc.py
hallgchris: Yeah in general that's probably true but since this is unlikely to ever be updated anyway, and the file is very minimal and straightforward to understand as it is, the comments are probably worth it to avoid future confusion.

hallgchris (approved):
I'm tired and apparently struggling to discuss BOOST with Angus right now, so I hope I'm not missing things I'd normally pick up :sweat_smile: but looks really good! This is shaping up really great. The simplicity of `main.py` shows how well this is put together :)

wireless_modules/libraries/abc.py
khanguslee: ohhhhhhh, I actually never bothered to look at what a decorator does. I just knew I had to add it before certain functions in python. This makes total sense now.

khanguslee (approved):
LGTM! lookin good now 🥇

wireless_modules/common_sensors/mq135_sensor.py
kdav108: I'm assuming by raw data you mean storing temperature and humidity separately in `dht_sensor`. I had originally done that but I felt that was only useful for this one method in a different class, which tbh I'm not even sure should use temperature and humidity values or not - the original source of the library I used for MQ135 mentions not using temp and humidity as they aren't as reliable yet, but I placed it in this class nevertheless in case the library does get updated to make use of these values better or we find that using them is better than not using them in terms of accuracy. You can find a link to the article on the notion page, but here it is as well (scroll to the bottom of the article: https://hackaday.io/project/3475-sniffing-trinket/log/12363-mq135-arduino-library )

hallgchris: Fair enough! We should do our own testing at some point to see if changing the temperature of the sensor affects the CO2 readings, should be fairly straightforward to do.

PR #63, opened 2020-09-18
Refactor ESP32 file upload script

kdav108:
## Description

In my previous PR (Middle Wireless Module sensor programming) Angus had suggested to refactor the `upload.sh` file since it contained a lot of repeated code and could become painful for maintaining going froward. So this PR refactored this shell script and also added a new functionality (why not).

Learning shell scripting wasn't as tough as I'd thought but it does follow some very different syntax (obviously) and so I tried to comment the important differences which I thought would make it easier for anyone else reading it and more so for myself if I ever come back to it.

The functions that `upload.sh` performs now include:
1) Uploading all the relevant files for middle wireless module to function onto the ESP32 board if used with only one argument (The port name)
2) If file paths are specified after the port name in the command line argument, the script will ONLY upload those files onto the board (This would make it faster when you only want to upload a few changed files rather than all files which is what the script did before) 

## Screenshots

Uploading all default files:
Screen Shot 2020-09-18 at 9 23 59 am

Uploading with wrong file path:
Screen Shot 2020-09-18 at 9 23 28 am

## Steps to Test

- Navigate to the `middle_module` folder
- Open terminal and try running `./upload.sh` : it shall fail and give you a friendly reminder
- Open a file used by the middle wireless module (e.g. `../libraries/abc.py`) and add some dummy code or comment at the bottom of the file (makes it easier to spot the change)
- Now type `./upload.sh <port_name>` replacing the placeholder with your port name for the ESP32 and this should upload all files onto the board and display the execution message.
- Alternatively type `./upload.sh <port_name> [files]` placing as many file paths you want updated in place of the [files] argument and the script should only upload those files.
- Use `ampy -p <port_name> -b 115200 get <file_changed>` replacing the relevant placeholders with the correct file names to view the changed files (if your change was towards the end of the file it should be easy to spot it)

### DIFF
--- a/wireless_modules/middle_module/upload.sh
+++ b/wireless_modules/middle_module/upload.sh
@@ -2,6 +2,14 @@
 
 # Note: Shell scripting should have no white space between the '=' operator for variable assigning
 
+# declares an integer
+declare -i BAUD_RATE=115200
+
+PORT=$1
+files="main.py ../config.py ../wireless_module.py  ../boot.py ../mqtt_client.py ../sensors/mpu.py \
+../sensors/dht_sensor.py ../sensors/co2_sensor.py ../libraries/MQ135/mq135.py ../sensors/sensor_base.py \
+../libraries/abc.py ../libraries/MPU6050-ESP8266-MicroPython/mpu6050.py"
+
 # Check that the port name is specified in the argument
 if [ $# -eq 0 ]
 then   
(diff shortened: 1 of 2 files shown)


wireless_modules/middle_module/upload.sh
```diff
@@ -1,25 +1,71 @@
...
+
+# NOTE: 
+# Shell scripting should have no white space between the '=' operator for variable assigning
+
+# declares an integer
+declare -i BAUD_RATE=115200
+
+PORT=$1
+NUM_ARGS=$#
+files="main.py ../config.py ../wireless_module.py  ../boot.py ../mqtt_client.py ../sensors/mpu.py \
```
hallgchris: Can this be stored in an array of some sort, ideally with one file per line? This will make git play much more nicely with this file, as it's likely that this file is frequently edited and likely to cause conflicts.

hallgchris (changes requested):
Not knowing much about shell script myself, this looks good!

I am thinking though, maybe we should at some point write an upload script in python... There's really no reason to use shell script and python would be easier to understand, extend, and is of course compatible with windows... But for now this is still a step forward!

Only other change I would suggest is the ability to call this script from any directory, e.g. by calling `middle_module/upload.sh /dev/ttyUSB0`. If this is too much work I'm happy to forgo this feature for now cos there's no point investing effort if we anticipate it being with a python script later.

wireless_modules/middle_module/upload.sh
kdav108: When I first searched I couldn't find array types in shell, but I looked again and apparently it is possible, so it shall be refactored!

kdav108:
> Only other change I would suggest is the ability to call this script from any directory, e.g. by calling `middle_module/upload.sh /dev/ttyUSB0`. If this is too much work I'm happy to forgo this feature for now cos there's no point investing effort if we anticipate it being with a python script later.

I think this is possible and I have somewhat of a clue but it would make this script much more complex and I actually like the idea of having a python script instead, so as you said, don't think it would be worth investing time on getting this extra feature done here.

hallgchris (approved):
Looks good to me! Nice quick PR :+1:

khanguslee:
> Only other change I would suggest is the ability to call this script from any directory,

Wouldn't you just add this script into your PATH?

khanguslee (approved):
wow. you just made your own arg parser too.

hallgchris:
> Wouldn't you just add this script into your PATH?

What I meant is, if you call this from another folder by either using an absolute path or by specifying a subfolder, all the relative paths to the python files are wrong, as they assume the current directory is the directory containing `upload.sh`.

kdav108:
> > Only other change I would suggest is the ability to call this script from any directory,
> 
> Wouldn't you just add this script into your PATH?

But the script would fail cos the files to load inside the script would then have the wrong path (cos they're defined as relative and not absolute path). It could work if we put each of the these files to load into the environment PATH, but not sure how effective/efficient that is.

PR #66, opened 2020-12-07
Update MQTT topics

hallgchris:
## Description

Fairly simple PR. This replaces all MQTT topics in the `das` module with the constants from our `mhp` module, and manually updates the topics in the Ant+ and WM code.

I think it would be possible to swap over the wireless module code to also use common repo topics, but it'll be more complicated so I'll leave that for another day.

This PR depends on monash-human-power/common#12. Once that PR is merged, I'll need to update `pyproject.toml`. Don't approve before this, otherwise I'll totally forget haha (pls still review though)

## Screenshots

n/a

## Steps to Test

Confirm that all scripts work as expected. Keep in mind that if you're testing them with other projects, you'll need to use the updated PRs for them too so that they MQTT topics match!


Blake-Haydon (commented):
This looks good, will approve once the other PR goes through. We might need to refactor the new logger as the one in the PR is old.

DAS/das/utils/DataToTempCSV.py
```diff
@@ -1,9 +1,23 @@
...
 import os
-import csv
 
-from das.utils import WirelessModule, WirelessModuleType
+from mhp.topics import WirelessModule
+
+
+@unique
+class WirelessModuleType(Enum):
+    """Used in the DataToTempCSV script to specify the type of data
```
nix-chu: Needs imperative mood here.

DAS/das/tests/fake_module_tester.py
```diff
@@ -163,22 +164,22 @@ def publish(client, topic, data={}):
...
     json_data = json.dumps(data)
 
     # Publish the data over MQTT
-    client.publish(topic, json_data)
+    client.publish(str(topic), json_data)
     print(topic, "--> ", json_data)
 
 
 def start_modules(args):
     """ Sends a null message on the start channels for all of the selected
```
nix-chu, line 179: Needs imperative mood here.

nix-chu (changes requested):

DAS/das/tests/fake_module_tester.py
hallgchris: Good catch! However, it seems that much of the documentation in these files (unchanged in this PR) aren't written in imperative mood, so I reckon it's possibly best left for another PR. @Blake-Haydon, perhaps this could be sneaked into your existing documentation PR? I promise to give you a speedy review.

DAS/das/tests/mqtt_test.py
```diff
@@ -29,19 +31,20 @@
 
 def start_publishing(client, args):
     print("start")
-    client.publish("power_model/start")
-    client.publish("start")
+    client.publish(str(BOOST.start))
```
kdav108: Why do we need to convert these to string, aren't they already a string?

DAS/das/tests/mqtt_test.py
```diff
@@ -2,6 +2,8 @@
 import argparse
 import paho.mqtt.client as mqtt
 
+from mhp.topics import DAS, BOOST
```
kdav108: Is It possible to import these and rename as `DASTopics` and `BOOSTTopics` or similar? It would make them more readable and avoid misunderstanding when reading lines such as `DAS.start` (vs `DASTopics.start`), in my humble opinion.

DAS/das/mqtt_wireless_logger.py
```diff
@@ -6,8 +6,9 @@
 import glob
 import re
 
+from mhp.topics import WirelessModule
```
kdav108: Same concern here, is it possible to import and rename to `WirelessModuleTopics`, or I think it may be a good idea to rename the class in `mhp.topics` directly. 
We also have a `WirelessModule` class in `wireless_module.py` and so having two classes with the same name can become confusing going forward, especially given that we would likely need to import `WirelessModule` from `mhp.topics` in the `wireless_module.py` file in the future to remove the literal topic strings used in that file.

kdav108 (changes requested):
Nice straightforward PR! We can finally have a more robust and maintainable code, at least in terms of changing and updating mqtt topics.

DAS/das/tests/mqtt_test.py
hallgchris: They're not, they're actually our `Topic` class. Having a class means we can do fancy stuff to build new Topics and compare them but unfortunately we do at some point need to convert to a string.

DAS/das/tests/mqtt_test.py
hallgchris: You've a very good point - I've had a think about this, and I totally agree withe that premise but I think lengthening the names would be redundant as the fact that it is a topic is made clear from the namespace. If we wanted to make the meaning of these more clear, we could instead import `from mhp import topics` and then call `topics.DAS.start` which is just as descriptive.

The question then becomes whether we should *actually* import like this instead. I feel like many would be resistant to it just because of the verbosity. I do agree with the points you've made (especially with regard to the naming collision with `WirelessModule` so I'll at least change it over in this repo.

DAS/das/tests/mqtt_test.py
kdav108: wait so the `start` attribute in `BOOST` is a class?

DAS/das/tests/mqtt_test.py
kdav108: Tbh I actually like the idea of using `topics.DAS.start`. I don't personally think it's too verbose, it actually helps make the code more readable, for me at least.

DAS/das/tests/mqtt_test.py
hallgchris: Sorry probably should have said! That was one of the changes in monash-human-power/common#12. See [here](https://github.com/monash-human-power/common/blob/6dd9bd5bcb8395f6e88b6364e568884fc8eb18a1/mhp/topics/topics.py#L60).

kdav108: Oh I see, interesting implementation!

hallgchris:
Common PR is in, ready to merge if you're all happy!

nix-chu (approved):
LGTM!

haotongwang (approved):

kdav108 (approved):
👌

PR #69, opened 2021-01-09
Add Battery voltage reader class

kdav108:
## Description

This pull request adds the `BatteryReader` class to the WM's arsenal to read the battery voltage powering the WM. The other relevant files have been updated to reflect this new entry. 

This makes our middle WM (or at least the MVP) complete in that it can carry out all the significant tasks required for it to conduct including reading data from the CO2, accelerometer, gyroscope, temperature and humidity sensors and reading battery voltage and sending them wirelessly using `mqtt`.

## Screenshots
Note I didn't have a battery connected in this case:
Screen Shot 2021-01-09 at 6 27 52 pm

## Steps to Test

1) Connect ESP32 to your computer or to test the battery voltage algorithm only use a battery to power the board.
2) The board should automatically run `boot.py` and then `main.py`
3) Use terminal window and enter `mosquitto_pub -h <broker name> -t /v3/wireless_module/2/start -m ""` to inform ESP32 to start sending data (Remember your `broker name` should match the one in your `config.py`)
4) Use `mosquitto_sub -h <broker name> -t /v3/wireless_module/2/data` to read sensor data and ensure they run correctly as they did before this PR.
5) Use `mosquitto_sub -h <broker name> -t /v3/wireless_module/2/battery` to read battery voltage. Remember that it only publishes battery data every 5 minutes. If needed for testing, you can change this rate in `main.py` or `wireless_module.py`.
6) Check that the data published confirms with your own reading using a voltmeter.


wireless_modules/wireless_module.py
```diff
@@ -14,16 +14,18 @@ class WirelessModule:
...
         Initialises the wireless module.
         :param module_id: An integer representing the wireless module number.
+        :param battery_reader: A `BatteryReader` class to read the battery voltage from
         """
         self.sensors = []
 
-        self.pub_data_topic = b"/v3/wireless_module/{}/data".format(module_id)
+        self.pub_sensor_topic = b"/v3/wireless_module/{}/data".format(module_id)
+        # FIXME: Need to verify the need for the low-battery topic
         self.pub_low_battery = b"/v3/wireless_module/{}/low-battery".format(module_id)
```
kdav108: Need to have a look at whether we need such a battery topic.

kdav108 (commented):
Do you guys think we should have separate `mqtt` topics for starting the sensor data publication and starting the battery data publication? I don't think it's necessary but just putting it out there if anyone thinks it may be useful.

wireless_modules/middle_module/main.py
```diff
@@ -23,8 +29,11 @@
...
 my_mq135.set_rzero(RZERO)
 
+battery_pin = 33
+battery_reader = BatteryReader(battery_pin, VOLTAGE_FACTOR)
+
 # Set up the wireless module
-middle_module = WirelessModule(MODULE_NUM)
+middle_module = WirelessModule(MODULE_NUM, battery_reader)
 sensors = [my_mpu, my_dht, my_mq135]
 middle_module.add_sensors(sensors)
```
kdav108, line 39: I know this is bad programming practice, but should we wrap the `middle_module.run()` statement on `line 42` (sorry git couldn't let me highlight it) inside a broad `try` and `except` block as a safety net in case something we had no idea of causes our system to fail on the race day. And we could simply reboot the system again, which in most cases should solve the issue (especially for this kind of systems). 
Let me know what you guys think.

wireless_modules/README.md
```diff
@@ -31,12 +31,23 @@ You will need to create a local version of `config.py` (`cp src/config.example.p
...
+The `upload.sh` file in `middle_module` folder can be used to upload all relevant files on the ESP32.
 ```
 # Change this to the serial port of the ESP32
 ./upload.sh --port /dev/ttyUSB0
 ```
 
+## Running code on the ESP32
+If using `picocom`:
+1) Open terminal and connect the ESP32 to your computer
+2) Run `picocom <insert_device_port_number> -b 115200`. This would run `boot.py` and then `main.py`.
```
hallgchris, line 43: For me this doesn't actually cause these files to run, I have to actually reset the ESP32. By default I'm put into the python REPL

wireless_modules/README.md
```diff
@@ -31,12 +31,23 @@ You will need to create a local version of `config.py` (`cp src/config.example.p
...
+3) Use `mosquitto` to interact with ESP32 (see [this](https://github.com/monash-human-power/data-acquisition-system/tree/master/wireless_modules/middle_module#steps-to-test))
+4) If you find that the board is not publishing to the `mqtt` topic or is not running, try the next few steps.
+5) Use `Control-C` to open REPL
+6) To explicitly run a file, say `main.py` on the board use `import main.py`.
```
hallgchris: This should be `import main` I believe

wireless_modules/README.md
```diff
@@ -31,12 +31,23 @@ You will need to create a local version of `config.py` (`cp src/config.example.p
...
+
+Use `Control-A` and then `Control-B` to terminate `picocom`.
```
hallgchris: `Control-A` + `Control-X` is exit for me

wireless_modules/battery_reader.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        # Resistor values in the voltage divider design here:
+        # https://www.notion.so/Preliminary-Design-Report-2d0a06e271614ae886e7a2b3f88f93aa
+        self.R1 = 33.2
+        self.R2 = 100
+
+        # The factor to multiply the voltage at the battery pin with to get the battery voltage
+        self.voltage_factor = voltage_factor
+        if voltage_factor is None:
+            self.voltage_factor = (self.R1 + self.R2) / self.R1
```
hallgchris, line 25: I feel like we should require and always use the voltage factor. That way, the user can use their own resistor values, and possibly apply their own scalar on top of that (which in my experimentation was required)

wireless_modules/middle_module/README.md
```diff
@@ -45,5 +45,8 @@ This repository contains all the code related to Monash Human Power's middle wir
 4. If using `mosquitto`, use 
     - `mosquitto_pub -h <broker name> -t /v3/wireless_module/2/start -m ""` to inform the module to start sending data
     - `mosquitto_pub -h <broker name> -t /v3/wireless_module/2/stop -m ""` to inform the module to stop sending data
-    - `mosquitto_sub -h <broker name> -t /v3/wireless_module/2/data` to read data sent from the module
+    - `mosquitto_sub -h <broker name> -t /v3/wireless_module/2/data` to read sensor data sent from the module
+    - `mosquitto_sub -h <broker name> -t /v3/wireless_module/2/battery` to read battery data sent from the module, 
+    remember that battery is only sent every 5 minutes. Change this in `../wireless_module.py` or `main.py` if you want
+    to test it with faster data rate.
```
hallgchris: I think this could be more explicit with how to make the change: Suggest adding `battery_data_rate=1` to the `run()` call.

wireless_modules/middle_module/main.py
hallgchris: That sounds like a good idea to me! In fact, I reckon you could put the whole code in this safety net. If initialising fails a reset is unlikely to fix it, but it's the best we can do and might help. Maybe put a 5 second delay before resetting and print a message also.

wireless_modules/middle_module/test.py
```diff
@@ -25,9 +26,18 @@ def display_difference(old, new):
...
 
 def on_msg_receive(client, data, message):
+    """
+    You can find out more details about the parameters of this function at https://pypi.org/project/paho-mqtt/#callbacks
+    """
     # Extract data from mqtt topic
     data = message.payload.decode('utf-8')
     data_dict = json.loads(data)
+
+    if message.topic == SUB_BATTERY_TOPIC:
```
hallgchris: A neater solution would be to use
```python
from paho.mqtt import topic_matches_sub
if topic_matches_sub("/v3/wireless_module/+/battery", message.topic):
```
Then we wouldn't need the mutating global variable.

wireless_modules/middle_module/test.py
```diff
@@ -69,13 +79,17 @@ def on_msg_receive(client, data, message):
     STOP_TOPIC = f'/v3/wireless_module/{MODULE_NUM}/stop'
     SUB_DATA_TOPIC = f'/v3/wireless_module/{MODULE_NUM}/data'
 
+    SUB_BATTERY_TOPIC = f"/v3/wireless-module/{MODULE_NUM}/battery"
```
hallgchris: Should be `wireless_module`

wireless_modules/wireless_module.py
hallgchris: I think we've decided that there's no need for the low battery topic.

wireless_modules/wireless_module.py
```diff
@@ -84,19 +89,26 @@ def run(self, data_rate=1):
...
+            prev_battery_read = time.ticks_ms()
 
             while self.start_publish:
-                data = self._read_sensors()
-                print("-------Publishing--------")
-
-                # compute time difference
-                time_taken = time.ticks_diff(time.ticks_ms(), start) * ms_to_sec
-
+                # Publish the battery voltage of this wireless module if 5 minutes (300 seconds) have elapsed since last
```
hallgchris: Rather than hardcoding this number into the docs, refer to `battery_data_rate`.

wireless_modules/wireless_module.py
```diff
@@ -84,19 +89,26 @@ def run(self, data_rate=1):
...
+                # battery read
+                time_since_last_battery_read = time.ticks_diff(time.ticks_ms(), prev_battery_read) * ms_to_sec
+                if time_since_last_battery_read >= battery_data_rate and self.battery is not None:
```
hallgchris, line 99: Can you make this always run when the module starts up? Probably as simple as initialising `prev_battery_read` to a time that has already expired. May as well do the same for data too, doesn't hurt

wireless_modules/wireless_module.py
```diff
@@ -84,19 +89,26 @@ def run(self, data_rate=1):
...
+                    battery_voltage = self.battery.read()
+                    self.mqtt.publish(self.battery_topic, ujson.dumps(battery_voltage))
```
hallgchris: Set `retain=True` so that if we want to check the battery level via MQTT, we don't have to wait until the next publication when subscribing

hallgchris (changes requested):
A few changes but will be awesome to have this finished!

> Do you guys think we should have separate mqtt topics for starting the sensor data publication and starting the battery data publication?

Good question, I think we should always be publishing battery data. The battery level is still important when the bike isn't moving!

wireless_modules/README.md
kdav108: `import main` doesn't seem to work when I try it. And interestingly (I just noticed this now), `import main.py` works only the first time I use it, and then it throws an `ImportError`. So I have to close picocom and start again.
Screen Shot 2021-01-13 at 9 07 05 pm

wireless_modules/README.md
kdav108: So what you see by default is `>>>`? It works for me (the files run right after I use the `picocom` command), but from my experience I noted that this method only runs `boot.py` and `main.py` when you first use `picocom` after connecting the ESP32, after that you have to go into Python REPL and run `import main.py`. 
Screen Shot 2021-01-13 at 9 05 27 pm

wireless_modules/README.md
kdav108: Maybe I'll document this in the `README.md`

wireless_modules/README.md
kdav108: My bad, that was a typo, sorry if that lead to a lot of web surfing for you

wireless_modules/battery_reader.py
kdav108: From my conversation with Brenny (@bdinh123), we're planning to make the same voltage divider (i.e same resistor values) for each of the WMs. Not sure if that's changed now but if not, then having a default `voltage_factor` would allow the three WM `main.py` scripts to not need to care about the circuit logic and design. And if the user needs to specify separate resistor values, they still can do that through passing their own `voltage_factor` argument.

wireless_modules/middle_module/README.md
kdav108: good point!

wireless_modules/README.md
kdav108: Correction: Just tried it again today, and `picocom` command automatically ran the two files in multiple occasions, so it's not that it only works the first time you use it after connecting the ESP32 board. Maybe there's something else going on which I can't sniff out.

hallgchris: That's weird. For me, `boot.py` and `main.py` will automatically run on power up i.e. when it's plugged in, but running `picocom` doesn't actually do anything.

So if you
- Plug in the ESP32
- Run `picocom`, and stop the program with control+c
- Exit `picocom`
- Run `picocom` again

The program runs automatically for you?

wireless_modules/README.md
hallgchris: This is weird, `import main.py` only works on one of my ESP32s. I know one of them has newer firmware than the other, perhaps that's what's causing the discreprancies

wireless_modules/README.md
hallgchris: Haha no none at all :)

wireless_modules/battery_reader.py
hallgchris: Fair point! In that case I think we just need a bit more flexibility. In my testing, the resistor values on their own don't give a perfect reading, I had to scale my voltages up a few percent in order to be accurate (probably something to do with the ADC). So we probably need something like the ability to scale the calculations from the resistors, rather than only the ability to override entirely.

wireless_modules/middle_module/test.py
```diff
@@ -25,9 +26,18 @@ def display_difference(old, new):
...
 def on_msg_receive(client, data, message):
+    """
+    You can find out more details about the parameters of this function at https://pypi.org/project/paho-mqtt/#callbacks
+    """
     # Extract data from mqtt topic
     data = message.payload.decode('utf-8')
     data_dict = json.loads(data)
+
+    if message.topic == SUB_BATTERY_TOPIC:
+        print("Battery voltage received: " + str(data_dict["percentage"]))
```
hallgchris: I think this should be `data_dict["voltage"]` :)

wireless_modules/README.md
kdav108: Hmmm that could be it. Do we know which version of the firmware we'll use on the final bike?

wireless_modules/README.md
kdav108: Yep, but only in some circumstances which is weird. Like on one day it would run both files even the second time I run `picocom` and on other days it would only run the two files the first time I run `picocom`.

And the two files `boot.py` and `main.py` also run for me automatically when the ESp32 is powered up (when plugged in).

I might do some more testing with different combinations and see if I can spot a pattern.

wireless_modules/middle_module/main.py
kdav108: @hallgchris do you want to implement this in your current PR (Adding the GPS sensor module to back WM) since you made quite a few changes to `main.py` and this would avoid a lot conflicts for whoever merges their PR later.

wireless_modules/README.md
hallgchris: Probably if we had to standardise on one version, whatever is latest. Right now that's `esp32-idf4-20200902-v1.13.bin`

wireless_modules/middle_module/main.py
hallgchris: Sure thing!

wireless_modules/battery_reader.py
kdav108: Do you think the scaling factor you used would be constant across all ESP32s? If so we can just have that inside the class. If not, we can let that be an argument to `__init__` - there's only 2 arguments at the moment so I think this approach should be fine but I feel like there's probably a better solution I can't think of right now.

wireless_modules/README.md
kdav108: Alright, I've added both `import main.py` and `import main` to the instructions and included a small note about using one or the other depending on your ESP32 version.

kdav108: very strangely, `import main.py` isn't working all of a sudden when I try it on your back WM branch

wireless_modules/battery_reader.py
hallgchris: I'm not sure tbh... I think I'd rather have it as an argument though. It's the sort of thing we might want to change at short notice and this might make it easier!

wireless_modules/wireless_module.py
```diff
@@ -83,20 +88,27 @@ def run(self, data_rate=1):
             print("waiting for message")
             self.mqtt.wait_for_message()
 
-            # get millisecond counter
-            start = time.ticks_ms()
-
+            # get millisecond counter and initialise to some previous time to start data publication immediately
+            prev_data_sent = time.ticks_ms() - 1000
```
hallgchris: Just for neatness, maybe subtract `data_rate` like you did for battery

wireless_modules/battery_reader.py
hallgchris: I think the best solution would be to have `voltage_factor` default to 1, and multiply the resistor-based calculation here by the voltage factor

hallgchris (changes requested):
Nearly done!

wireless_modules/battery_reader.py
kdav108: But that would mean the resistor values won't be easily modifiable. And a scenario where we decide to use different resistors for a given WM can occur, so maybe we should have both the resistor-based calculation and the scaling factor as an argument?

hallgchris: Sorry, I didn't see this reply earlier somehow. I think that would be ideal! We want to avoid having to touch anything except `main.py` when running these.

hallgchris (changes requested):
Just the one thing

kdav108:
Added in the scale factor to `BatteryReader`. Atm it's defaulted to 1, but if we can figure out if there is one scaling factor that helps in most situations we can use that as the default. @hallgchris see if you can utilise the extra argument and get the correct reading without having to add anything to `main.py` apart from changing existing variables.

hallgchris (approved):
All looks good to me!

PR #70, opened 2021-01-12
Add GPS sensor to back WM

hallgchris:
## Description

This PR:
- Creates the back WM script, mainly copied off the middle WM code
- Switches to using `asyncio` for anywhere where the code has to wait for something to happen. This allows other code to run instead. (i.e. read data from UART while waiting for MQTT messages to come in)
- Implements the GPS sensor class.
  - I added the `datetime` and [`pdop`](https://en.wikipedia.org/wiki/Dilution_of_precision_(navigation)) fields which aren't on the V3 MQTT spec right now - I reckon they'll be good to have though.

## Screenshots

Here's some example generated data. (Location truncated for privacy as this is a public repo, normally shows to 7 sig figs)

```json
{
  "type": "gps",
  "value": {
    "satellites": 4,
    "pdop": 6.86,
    "latitude": -37.8,
    "longitude": 145,
    "altitude": 140.7,
    "speed": 0.7798978,
    "course": 13.69,
    "datetime": "2021-01-12T12:13:33",
  },
}
```

## Steps to Test

If you have an ESP32 and a compatible GPS module, you should be able to just
- Set the UART channel correctly in `main.py` for however your module is hooked up
- Run `main.py`
- Then, ensure reasonable information is published.


wireless_modules/back_module/upload.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
+
+# Purpose: To upload specific files or all relevant middle wireless module files (default) onto the ESP32
```
kdav108: `middle` -> `back`

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,60 @@
+from machine import UART
+import uasyncio as asyncio
```
kdav108, line 2: That's a smart idea, makes it seem like it's just normal python

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,60 @@
...
+        :param uart_channel: The UART channel that the GPS is connected to.
+        """
+        self.gps = MicropyGPS(location_formatting="dd")
+        asyncio.create_task(self.uart_rx(uart_channel))
+
+    async def uart_rx(self, uart_channel: int):
+        """
+        Read and load data from the GPS over UART.
+        """
+        uart = UART(uart_channel, 9600)
```
kdav108: Does the `9600` have some important meaning? If so, it may be helpful to turn the hardcoded value into a variable.

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,60 @@
...
+        return [
+            {
+                "type": "gps",
+                "value": {
+                    "satellites": self.gps.satellites_in_use,
+                    "pdop": self.gps.pdop,
+                    "latitude": latitude,
+                    "longitude": longitude,
+                    "altitude": self.gps.altitude,
+                    "speed": self.gps.speed[2] / 3.6,
```
kdav108: hardcoded value

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,60 @@
...
+                    "course": self.gps.course,
```
kdav108, line 59: Might help defining what `course` and `pdop` represent, perhaps in the method documentation. 
I couldn't find them on notion, but if it's there somewhere, we can alternatively just reference the link to notion in the method doc.

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Get the most recently stored GPS data.
+        :return: An array containing the GPS sensor dictionary. `value` in this
+        dictionary contains spacial, quality of service and time information.
+        """
+        latitude = self.gps.latitude[0] * (1 if self.gps.latitude[1] == "N" else -1)
+        longitude = self.gps.longitude[0] * (1 if self.gps.longitude[1] == "E" else -1)
+        # ISO 8601 datetime format
+        datetime = "20{:02d}-{:02d}-{:02d}T{:02d}:{:02d}:{:02.0f}".format(
+            self.gps.date[2], self.gps.date[1], self.gps.date[0], *self.gps.timestamp
```
kdav108, line 45: Just out of interest, why are we collecting date and time information?

wireless_modules/middle_module/main.py
```diff
@@ -28,6 +28,13 @@
...
-# Enters an infinite loop
-print("starting loop")
-middle_module.run()
+
+async def main():
+    print("Starting asyncio loop")
+    asyncio.create_task(middle_module.run())
+
+    while True:
+        await asyncio.sleep(1)
```
kdav108, line 50: Might help explain what's happening here with in-line comments, since to someone who's new to `asyncio` (me for example) this code looks redundant. Don't worry about it if it's supposed to be obvious to someone with basic knowledge of `asyncio`, most programmers on this repo will learn it.

wireless_modules/back_module/main.py
```diff
@@ -0,0 +1,37 @@
...
+
+async def main():
+    print("Starting asyncio loop")
+    asyncio.create_task(back_module.run())
+
+    while True:
+        await asyncio.sleep(1)
+
+
+asyncio.run(main())
```
kdav108: This is a bit pedantic, but just wanted to point this out in case in our repos we care about having a new line at the end of the file.

kdav108 (changes requested):
Great work, code _looks_ functionally correct (couldn't test it with a GPS module, but I trust it works smoothly). Nice twist in the WM story with the 'revolutionary' `asyncio` library 😂. Only learnt about concurrency, but this is the first time I'm looking at it in a proper implementation.

I just have a few suggestions, nothing too big.

wireless_modules/sensors/gps_sensor.py
hallgchris: Not my smart idea, copied from a `uasyncio` tutorial haha

wireless_modules/sensors/gps_sensor.py
hallgchris: It does! Will add that in, but for your interest it's the [baud rate](https://learn.sparkfun.com/tutorials/serial-communication/all#rules-of-serial) for the UART device

wireless_modules/sensors/gps_sensor.py
hallgchris: Not necessarily for a particular reason, but I thought it had potential to be useful in the future. The biggest use I see it potentially having is so we can work out when a log file is from, as the files are not saved with any meaningful timestamp.

wireless_modules/sensors/gps_sensor.py
hallgchris: Documentation added! https://www.notion.so/V3-MQTT-Topics-66e6715d0e1941ffaa82020e5868fbae#c56956ba5d384bc0ac246fd7e964a03b

wireless_modules/sensors/gps_sensor.py
kdav108: But then do you think we need to publish it every second? We already have the traveled distance data so we can  graph anything over time.

kdav108: But I guess it doesn't cost us much to publish it every second, so if we think it's alright, I guess it's fine!

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,61 @@
...
+                self.gps.update(chr(char))
+
+    def read(self):
+        """
+        Get the most recently stored GPS data.
+        :return: An array containing the GPS sensor dictionary. `value` in this
+        dictionary contains spacial, quality of service and time information.
+        """
+        latitude = self.gps.latitude[0] * (1 if self.gps.latitude[1] == "N" else -1)
+        longitude = self.gps.longitude[0] * (1 if self.gps.longitude[1] == "E" else -1)
```
Blake-Haydon, line 42: I'm assuming that the N and E correspond to north and east. Is there any chance you could add an inline comment to clear this up?

wireless_modules/sensors/gps_sensor.py
Blake-Haydon: @hallgchris collecting date and time from the GPS would be very handy for logging. I think it would solve the problem with not having a real time clock.

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,61 @@
...
+        # ISO 8601 datetime format
+        datetime = "20{:02d}-{:02d}-{:02d}T{:02d}:{:02d}:{:02.0f}".format(
+            self.gps.date[2], self.gps.date[1], self.gps.date[0], *self.gps.timestamp
+        )
+        kph_per_mps = 3.6
```
Blake-Haydon: I know this is a small nit pick but could we instead use `kph_to_mps = 1/3.6`. Then we can have `self.gps.speed[2] * kph_to_mps`. In my brain this makes more sense, but up to you.

Blake-Haydon (changes requested):

Blake-Haydon:
Oh also it looks really good, bit of a bummer I can't test it without the GPS module.

wireless_modules/back_module/main.py
```diff
@@ -0,0 +1,46 @@
+import time
+import uasyncio as asyncio
```
kdav108, line 3: I realised that these non-inbuilt microPython functions need to be manually installed using upip. We probably don't want to put code on the ESP32 and forget to install the right packages before plugging it onto a module and testing it. So shall we note this down somewhere in the README.md, or is it quite straightforward?

wireless_modules/back_module/main.py
```diff
@@ -0,0 +1,46 @@
...
+    # Keep our code running indefinitely - the above call won't block!
+    while True:
+        await asyncio.sleep(1)
+
+
+try:
+    asyncio.run(main())
+except KeyboardInterrupt:
+    print("KeyboardInterrupt, exiting...")
+except:
```
kdav108: Haven't used python `except` in a while, but I believe you can `except Exception as e:` and `print(type(e).__name__)`. This might help us know what the error was and help us debug. Hopefully Micropython supports all this.

kdav108 (changes requested):
Just a couple of more things I noted a bit late.

wireless_modules/back_module/main.py
hallgchris: `uasyncio` is part of micropython, no installation required!

wireless_modules/back_module/main.py
hallgchris: Very good point! Otherwise we'd never know what went wrong...

wireless_modules/back_module/main.py
kdav108: That's strange, when I first tried running the code it said `no uasyncio module found` or something similar, than I used pip to install it and it then worked. Maybe it's pre-installed for the newer version of ESP32?

hallgchris: As discussed, version 1.13 is required for `uasyncio`. I've added this info to the readme!

wireless_modules/sensors/gps_sensor.py
```diff
@@ -0,0 +1,63 @@
...
+
+class GpsSensor(Sensor):
+    """
+    A GPS sensor implementing NMEA-0183 connected via UART, such as the u-blox NEO 6M.
+    """
+
+    def __init__(self, uart_channel: int):
+        """
+        Begin reading data from the GPS over UART.
+        :param uart_channel: The UART channel that the GPS is connected to.
```
kdav108, line 17: Is this different to a pin on ESP32?

kdav108, line 17: Is this different to a pin on the ESP32?

kdav108 (dismissed):
I just have one comment (for my own understanding), apart from that this PR looks good to go through!

wireless_modules/sensors/gps_sensor.py
hallgchris: Yep! The ESP32 has 3 UART controllers, each is hardwired to a pair of pins. More about UART [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html), or you can see 2 of the UART channels available on this pin diagram below. (not sure where UART1 is... I'm wondering if it's connected to USB on the devboard or something like that)
image

hallgchris:
Merged in the battery voltage reading code, a bit of a merge conflict but all fixed up.

@Blake-Haydon can you re-review when you get the chance? Your old review is blocking :)

wireless_modules/middle_module/main.py
```diff
@@ -16,27 +19,44 @@
...
+    sda_pin = machine.Pin(21)
+    dht_pin = machine.Pin(4)
+    mq135_pin = machine.Pin(34)
+
+    # Instantiate sensor objects
+    # my_mpu = Mpu(scl_pin, sda_pin, 20)
+    # my_mpu.calibrate()
+    # my_dht = DhtSensor(dht_pin)
+    # my_mq135 = CO2(mq135_pin)
+    # my_mq135.set_rzero(RZERO)
```
kdav108: Just one small thing - I believe you forgot to uncomment these after testing 😂. Same goes for `line 42` below.

Blake-Haydon (dismissed):
Looks good! Already checked beforehand.

wireless_modules/middle_module/main.py
hallgchris: Wow, thank you for picking this up haha

hallgchris:
Sorry just need one more review, made a silly mistake in the merge :)

kdav108 (approved):
🏁

PR #72, opened 2021-01-25
Add reed switch sensor

hallgchris:
## Description

This PR adds the reed switch speed/distance sensor!

Note I've set the GPS module branch as the base for this merge. I'll change it to master once that's in, but this way you can see and review the changes without having all the GPS-related changes messing up the diffs.

@maanika - Would be good to get a review from you since you've been writing very similar code as part of your research project!

## Screenshots

Sample sensor output over MQTT. Mounted on a real bike wheel by the way :)
image
IMG_20210125_162650

## Steps to Test

All you need to test this is an ESP32 and pushbutton/switch! Load up the back module code onto it (removing the other sensors so you're not looking for a gps that isn't there) and wire up a push button from GPIO 5 to ground. Run the code, and each press of the push button will be interpreted as one wheel revolution.


wireless_modules/sensors/reed_sensor.py
```diff
@@ -0,0 +1,70 @@
+import machine
+import time
+
+from sensor_base import Sensor
+
+US_TO_S = 1e-6
+
+# Ignore anything shorter than 50ms, likely to be switch bouncing.
+# (roughly corresponds to 150 km/h)
+MIN_REVOLUTION_TIME = 50_000  # us
```
kdav108: Probably better to name it micro-second, `us` is a bit ambiguous.

wireless_modules/sensors/reed_sensor.py
```diff
@@ -0,0 +1,70 @@
...
+        :return: An array containing the velocity data and the distance data as
+            separate elements.
+        """
+        now = time.ticks_us()
+        diff = time.ticks_diff(now, self.last_trigger_time)
+        current_speed = self.current_speed if diff < MAX_REVOLUTION_TIME else 0
+
+        return [
+            {"type": "reedVelocity", "value": current_speed},
+            {"type": "reedDistance", "value": self.distance_travelled},
```
kdav108, line 63: If I understand correct, these values will be somewhat 'delayed' if the `read` method is called after `reed_callback`?

wireless_modules/wireless_module.py
```diff
@@ -72,12 +72,21 @@ def sub_cb(self, topic, msg):
...
-            self.mqtt.check_for_message()
-            await asyncio.sleep_ms(100)
+
+            while not self.start_publish:
+                self.mqtt.check_for_message()
+                await asyncio.sleep_ms(100)
+
+            # Start message received, tell sensors to start
+            for sensor in self.sensors:
+                sensor.start()
```
kdav108: Should be named `on_start` I believe

wireless_modules/sensors/reed_sensor.py
```diff
@@ -0,0 +1,70 @@
...
+        ]
+
+    def start(self):
```
kdav108: `on_start` perhaps?

kdav108 (changes requested):
Almost next to ideal! Just a few comments and questions.

wireless_modules/sensors/reed_sensor.py
hallgchris: That's completely correct, but nothing to worry about. At speed, `reed_callback` will be called around every 80 ms or so, so the data will never be very old. (The GPS sensor works similarly, by the way.)

wireless_modules/wireless_module.py
hallgchris: Good catch!

kdav108 (dismissed):
Looks 👌

hallgchris:
Merged in battery voltage and GPS sensor changes.

wireless_modules/back_module/main.py
```diff
@@ -13,24 +15,36 @@
...
     # Set up the wireless module
-    back_module = WirelessModule(MODULE_NUM)
-    sensors = [my_mq135, my_gps]
+    back_module = WirelessModule(MODULE_NUM, battery_reader)
+    sensors = [my_mq135, my_gps, my_reed]
     back_module.add_sensors(sensors)
 
     print("Starting asyncio loop")
-    asyncio.create_task(back_module.run())
+    asyncio.create_task(back_module.run(battery_data_rate=1))
```
kdav108: Do we want to change this to our intended `300` seconds instead.

kdav108 (dismissed):
Approved with just a small change we may want to make.

wireless_modules/back_module/main.py
hallgchris: Damnit that's me testing everythings working and not reverting changes again

hallgchris: Thanks for being such a diligent reviewer, fixed now

kdav108 (approved):
👌

PR #73, opened 2021-03-06
Update V3 fake module

haotongwang:
## Description

Updates the data sent to align with notion page

Made some values more sensible.

Distance increases.

Used to test [dashboard #107](https://github.com/monash-human-power/dashboard/pull/107)

## Steps to Test

1. Run `V3_fake_module` according to the DAS repo instructions.

### DIFF
--- a/DAS/das/utils/MockSensor.py
+++ b/DAS/das/utils/MockSensor.py
@@ -4,25 +4,31 @@ import random
 class MockSensor:
     """ Base class to make a mock sensor that produces random data"""
 
-    def __init__(self, *average_value, percent_range=0.05, decimals=2):
+    def __init__(self, *average_value, percent_range=0.05, decimals=2, increment=False):
         """
         average_value:  Either a single value that sets the average value for
-                        the sensor or an array of subvals formatted as
-                        (sub_valueue_name, average_value). Where
-                        sub_valueue_name is the name that will be used in the
+                        the sensor or a tuple of tuple subvalues formatted as
+                        (sub_value_name, average_value). Where
+                        sub_value_name is the name that will be used in the
                         outputted dictionary and the average_value is similar
                         to a single value.
         percent_range:  Percent range for the randomly generated value
         decimals:       Number of decimal places for the random generated val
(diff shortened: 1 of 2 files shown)


DAS/das/utils/MockSensor.py
```diff
@@ -57,4 +66,8 @@ def gen_single_value(self, average_value):
         sensor_val += average_value * random.uniform(-self.percent_range,
                                                      self.percent_range)
         sensor_val = round(sensor_val, self.decimals)
+        if self.increment and self.single_val:
+            new_value = average_value + abs(average_value - sensor_val)
+            self.average_value = (new_value, )
+            return new_value
```
Blake-Haydon, line 72: Should this be `self.average_value = new_value` instead of `self.average_value = (new_value, )`?

Blake-Haydon (approved):
Can't test it on a real board but the changes look all good apart from the one weird line that was pointed out. It may be my mistake so I will give it a precarious approve.

DAS/das/utils/MockSensor.py
haotongwang: The type of `self.average_value` should be a tuple, `new_value` is a float. The `(new_value, )` syntax is to create a tuple with one element (which is what happens when there is a single value).

DAS/das/utils/MockSensor.py
```diff
@@ -4,25 +4,31 @@
...
-                        (sub_valueue_name, average_value). Where
-                        sub_valueue_name is the name that will be used in the
+                        the sensor or a tuple of tuple subvalues formatted as
+                        (sub_value_name, average_value). Where
+                        sub_value_name is the name that will be used in the
                         outputted dictionary and the average_value is similar
                         to a single value.
         percent_range:  Percent range for the randomly generated value
         decimals:       Number of decimal places for the random generated val
+        increment:      Whether to increase the value or generate a new value.
```
hallgchris, line 17: If you want to be super fancy, `monotonic` might be a good word to go with here

hallgchris (commented):
LGTM!

PR #77, opened 2021-08-23
Make random changes to run the WM code on our back WM

hallgchris:
## Description

Um basically just lots of random stuff. I'd list them here but honestly I'd just be repeating what's in the commit messages. Take a read below :)

One thing to note is I added some helpful scripts in a `utils` folder, not for deployment but they were useful in development and it seems like a waste to get rid of them.

LED status indication will come in a future PR.

## Screenshots

I'd screenshot the sensor output but it's late and I'm lazy. I'll add it tomorrow :)

Anyway here's some nice graphs from the utils.

Output from the battery calibration linear regression:

image

```
Slope:          0.7106689209693502
Intercept:      0.9267915203537527
R^2:            0.9847840762003852
```

Before and after adding the averaging of CO2 data. It was more noticeable in other runs, but notice how the line is more smooth after the gap in data. (The sudden increases are from breathing on the sensor)

co2

And finally here's what a full battery discharge looks like, with data logging and publishing on wifi. Overall it takes around 14.5 hours. Note the readings weren't calibrated well at this point, it actually began at around 4.1 V and finished around 3V.

image

## Steps to Test

Hard to test the hardware code without hardware, but all the `util` will run fine on your computer! I've tested everything on actual back WM, main thing is make sure I've not done anything stupid, and let me know if any code could be improved.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,7 +8,7 @@
 	ignore = untracked
 [submodule "wireless_modules/libraries/MQ135"]
 	path = wireless_modules/libraries/MQ135
-	url = https://github.com/rubfi/MQ135
+	url = https://github.com/monash-human-power/MQ135.git
 [submodule "wireless_modules/libraries/MPU6050-ESP8266-MicroPython"]
 	path = wireless_modules/libraries/MPU6050-ESP8266-MicroPython
 	url = https://github.com/adamjezek98/MPU6050-ESP8266-MicroPython
--- a/wireless_modules/back_module/main.py
+++ b/wireless_modules/back_module/main.py
@@ -26,7 +26,15 @@ def battery_calibration(voltage):
     return 0.710 * voltage + 0.927
 
 
+# Pins for R, G, B channels of status LED
+STATUS_PINS = [14, 12, 13]
+
+
 async def main():
+    for pin_num in STATUS_PINS:
+        pin = machine.Pin(pin_num, machine.Pin.OUT)
+        pin.off()
+
     # Define all the Pin objects for each sensor
     dht_pin = machine.Pin(4)
     reed_pin = machine.Pin(5)
--- a/wireless_modules/back_module/main.py
+++ b/wireless_modules/back_module/main.py
@@ -6,6 +6,7 @@ from wireless_module import WirelessModule
 from co2_sensor import CO2
 from dht_sensor import DhtSensor
 from gps_sensor import GpsSensor
+from mpu_sensor import MpuSensor
 from reed_sensor import ReedSensor
 from battery_reader import BatteryReader
 
@@ -27,8 +28,10 @@ SCALE_FACTOR = 4.13 / 4.64
 async def main():
     # Define all the Pin objects for each sensor
     dht_pin = machine.Pin(4)
-    mq135_pin = machine.Pin(34)
     reed_pin = machine.Pin(5)
+    sda_pin = machine.Pin(21)
+    scl_pin = machine.Pin(22)
+    mq135_pin = machine.Pin(34)
 
     # Instantiate sensor objects
     my_dht = DhtSensor(dht_pin)
@@ -40,6 +43,9 @@ async def main():
 
     my_gps = GpsSensor(2)
 
+    my_mpu = MpuSensor(scl_pin, sda_pin, 20)
+    my_mpu.calibrate()
+
     battery_pin = 32
     battery_reader = BatteryReader(
         battery_pin, scale=SCALE_FACTOR, voltage_factor=VOLTAGE_FACTOR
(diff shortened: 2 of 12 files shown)


.gitmodules
```diff
@@ -8,7 +8,7 @@
 	ignore = untracked
 [submodule "wireless_modules/libraries/MQ135"]
 	path = wireless_modules/libraries/MQ135
-	url = https://github.com/rubfi/MQ135
+	url = https://github.com/monash-human-power/MQ135.git
```
kdav108, line 11: You could open a PR on `rubfi/MQ135` with the improvements. Our `MQ135` repo doesn't seem to have much more change, so it should be quite easy, in theory, to get it into the source repository. That way we don't have to maintain it ourselves. Just a suggestion, won't impact this PR since getting our stuff in `rubfi/MQ135` might take a while.

wireless_modules/back_module/main.py
```diff
@@ -13,34 +15,52 @@
 MODULE_NUM = "3"
 
 # RZERO (for calibration of MQ135) found when the sensor was first 'activated'
-RZERO = 8.62
+RZERO = 6.1
```
kdav108: Can we document this somewhere (if not already done), because the R-zero might differ between CO2 sensor, and it might be good to record it somewhere so we don't need to re-calibrate and find it

wireless_modules/back_module/main.py
```diff
@@ -13,34 +15,52 @@
...
 
 # Circumference of a 700x28 bike wheel in meters.
 WHEEL_CIRCUMFERENCE = 2.136
 
-# Define Voltage divider factor for this module or leave as None to use default voltage factor
-VOLTAGE_FACTOR = None
+# Define a function to map measured voltages to "true" voltages.
+# See util/battery_lin_reg.py
+def battery_calibration(voltage):
+    return 0.710 * voltage + 0.927
```
kdav108, line 29: Hardcoded values

wireless_modules/battery_reader.py
```diff
@@ -3,7 +3,7 @@
 
 
 class BatteryReader(Sensor):
-    def __init__(self, pin_num=33, scale=1, voltage_factor=None):
+    def __init__(self, pin_num=33, calibration_func=lambda x: x):
```
kdav108, line 6: We forgot to update the func docs below for this replacement parameter

wireless_modules/battery_reader.py
```diff
@@ -17,18 +17,13 @@ def __init__(self, pin_num=33, scale=1, voltage_factor=None):
         self.R1 = 33.2
         self.R2 = 100
 
-        self.scale_factor = scale
-
-        # The factor to multiply the voltage at the battery pin with to get the battery voltage
-        self.voltage_factor = voltage_factor
-        if voltage_factor is None:
-            self.voltage_factor = (self.R1 + self.R2) / self.R1
+        self.calibration_func = calibration_func
```
kdav108, line 19: Like the idea of taking a func as parameter instead, gives more flexibility to whoever is using this class

kdav108 (changes requested):
Tried my best to "mentally" run the scripts 😂, but I trust it does work correctly :)

Just a few minor suggestions...

.gitmodules
hallgchris: I considered opening a PR, but decided against it for a few reasons. The original repo and its owner seem to be completely inactive, and haven't responded to existing PRs. That aside, this change would totally break any existing code so I'm not sure it's a great addition to the main library.

wireless_modules/back_module/main.py
hallgchris: Honestly I think it's okay here. Think of the entire function as a single constant value which I've extracted rather than the alternative of burying it in the battery code. I think these values are clearly a linear and constant term of some relation but naming variables like `BATTERY_CALIBRATION_LINEAR_TERM` wouldn't make things easier to read. I'll expand the documentation a bit to explain how they were generated more explicitly and state that it's just a linear relation.

Let me know your thoughts!

kdav108: Yep I agree, don't need to add heavy variable names here, but my concern with hardcoded values is how should one understand the meaning and origin of those values. I guess it is obvious that they're constant of a simple linear equation, but it might be nice to document how they were found, perhaps mention the `battery_lin_reg.py ` script.

wireless_modules/back_module/main.py
kdav108: Also, maybe I wasn't as clear initially, but should we document in notion perhaps, the `RZERO` value for each sensor we have calibrated. In case we end up using the sensor I used initially, we don't want to do the 24-hour heat up and measure this again. So it might be good to map what the `RZERO` value is for each sensor we have calibrated.

hallgchris: That's definitely a good idea!

For the time being I've reverted changing the RZERO in this PR, since I've realised my testing wasn't valid since I did it in my room rather than somewhere ventilated.

kdav108 (approved):
LGTM!

PR #78, opened 2021-08-30
Add systemd file for Mqtt logger

kdav108:
## Description

This PR adds the files needed to set up our own systemd service for the MQTT logger. It also makes a change to the `V3_fake_module.py` script to send a stop message whenever the script is terminated.

## Screenshots

After running `install.sh` you should see the `logger` in the user systemd folder:
Screen Shot 2021-08-30 at 12 07 22 pm

Logs from the `logger` service when data is being published and then stopped:
_Note: The printing of `/v3/wireless_module/3/data` was manually added just to check the logger was working correctly:_

Screen Shot 2021-08-30 at 12 06 47 pm

## Steps to Test

1) On your machine or Raspberry Pi, on the command line run `./install.sh` to set up the `logger` service
2) Check the systemd folder for user service files to confirm logger service is present there
3) Follow the README to enable and then start the service
4) Run the fake WM script and then stop it to see if the logger has successfully logged the data and create a csv file in the `das` folder.


kdav108:
Thanks for the great additions @Blake-Haydon!

Blake-Haydon (approved):
Works as intended. LGTM!

PR #79, opened 2021-09-06
Migrate to sqlite from CSV files

Blake-Haydon:
## Description

This swaps out the logging from using temporary CSV files to a more stable SQLite local database. 

Nothing too scary, just stores the MQTT messages in the database file called `MQTT_log.db`

## Screenshots

Screen Shot 2021-09-13 at 4 32 29 pm
Example end of log

## Steps to Test

These steps assume that you are in the data-acquisition-system root directory and poetry installed.
1. `cd DAS`
2. `poetry install`
3. `poetry shell`

**TERMINAL 1**
`python -m das.V3_mqtt_recorder "#" -v`

**TERMINAL 2**
`python -m das.V3_fake_module -t 10 -r 3`

**READ AND CHECK THE DATABASE**
You can read the database (`MQTT_log.db`) manually like a hacker 

OR...

Just use a GUI tool to inspect the data like the one below. 
Screen Shot 2021-09-13 at 4 35 03 pm

### DIFF
--- a/DAS/README.md
+++ b/DAS/README.md
@@ -9,7 +9,7 @@ These steps assume that you are in the data-acquisition-system root directory an
 1. `cd DAS`
 2. `poetry install`
 3. `poetry shell`
-4. Run any DAS command, _e.g._ `python -m das.V3_mqtt_recorder`
+4. Run any DAS command, _e.g._ `python -m das.V3_mqtt_recorder -v`
 
 ## Run Tests
 These steps assume that you have done the basic setup.
@@ -47,13 +47,13 @@ This command line tool reads a raw csv log created from the V3 MQTT Recorder too
 ### Usage
 ```
 # General command
-python -m das.V3_mqtt_playback [FILEPATH] [FLAGS]
+python -m das.V3_mqtt_playback [FLAGS]
 
-# Playback of 1_log.csv
-python -m das.V3_mqtt_playback ./das/csv_data/1_log.csv  -v
+# Playback of all logs
+python -m das.V3_mqtt_playback -v
 
-# Playback of 2_log.csv at 60x speed 
-python -m das.V3_mqtt_playback ./das/csv_data/2_log.csv -s 60 -v
+# Playback of all logs at 60x speed 
+python -m das.V3_mqtt_playback -s 60 -v
 ```
 
 | Flag                          | Default Value |               Info               |
--- a/DAS/README.md
+++ b/DAS/README.md
@@ -37,6 +37,8 @@ python -m das.V3_mqtt_recorder /v3/wireless_module/# boost/# -v
 | `--host HOST`              |  `localhost`  |             Address of the MQTT broker              |
 | `-v ` or `--verbose`       |    `False`    |               Verbose logging output                |
 | `-t TIME` or `--time TIME` |     `inf`     | Length of time to record data (duration in seconds) |
+| `--username USERNAME`      |    `None`     |            Username for the MQTT broker             |
+| `--password PASSWORD`      |    `None`     |            Password for the MQTT broker             |
 | `-h` or `--help`           |               |                        Help                         |
 
 <br/>
@@ -61,6 +63,8 @@ python -m das.V3_mqtt_playback -s 60 -v
 | `--host HOST`                 |  `localhost`  |    Address of the MQTT broker    |
 | `-v ` or `--verbose`          |    `False`    |      Verbose logging output      |
 | `-s SPEED` or `--speed SPEED` |      `1`      | Playback speed up (x multiplier) |
+| `--username USERNAME`         |    `None`     |   Username for the MQTT broker   |
+| `--password PASSWORD`         |    `None`     |   Password for the MQTT broker   |
 | `-h` or `--help`              |               |               Help               |
 
 <br/>
--- a/DAS/README.md
+++ b/DAS/README.md
@@ -18,7 +18,7 @@ These steps assume that you have done the basic setup.
 <br/>
 
 ## [V3 MQTT Recorder](/DAS/das/V3_mqtt_recorder.py)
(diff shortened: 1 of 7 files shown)


DAS/README.md
```diff
@@ -37,30 +37,34 @@ python -m das.V3_mqtt_recorder /v3/wireless_module/# boost/# -v
 | `--host HOST`              |  `localhost`  |             Address of the MQTT broker              |
 | `-v ` or `--verbose`       |    `False`    |               Verbose logging output                |
 | `-t TIME` or `--time TIME` |     `inf`     | Length of time to record data (duration in seconds) |
+| `--username USERNAME`      |    `None`     |            Username for the MQTT broker             |
+| `--password PASSWORD`      |    `None`     |            Password for the MQTT broker             |
```
kdav108, line 41: This doesn't have to concern this PR, but I was thinking about not taking the password over command line and reading it from a file instead. 

These are a few reasons I thought of for this:
- If you pass a command through command line, then anyone on the system can use `ps` to see the value
- You can see the pwd by using `history`
- Files can be encrypted/restricted to a particular user and so no one else can read it
- With Files we don't need to pass around the password between everyone working on this

Thought none of this concerns relate to the scale at which we work, it's good to ponder over and learn from these projects.

What are your thoughts?

DAS/das/utils/logger.py
```diff
@@ -24,54 +14,82 @@ class Recorder:
...
+        ).fetchone()
+
+        # Make a LOG table if none currently exists
+        if contains_LOG_table is None:
+            self._CONN.execute(
+                """
+                CREATE TABLE LOG
+                (TIME_DELTA DECIMAL(15,6)   PRIMARY KEY         NOT NULL,
+                TOPIC                       VARCHAR(250)        NOT NULL,
+                MESSAGE                     VARCHAR(500)        NOT NULL);
```
kdav108: Is this length tested so none of our messages get cut off (especially WM 3 data payload). Alternatively, we can leave it as just `VARCHAR` hence restricting sql from cutting off any message.

DAS/das/utils/logger.py
```diff
@@ -24,54 +14,82 @@ class Recorder:
...
+            """
+            SELECT name FROM sqlite_master WHERE type='table' AND name='LOG'
+            """
+        ).fetchone()
+
+        # Make a LOG table if none currently exists
+        if contains_LOG_table is None:
+            self._CONN.execute(
+                """
+                CREATE TABLE LOG
```
kdav108: Nit: might be nice if we use a variable for this table name, it will make it much easier and cleaner in my opinion if we ever decide to change it.

DAS/das/utils/logger.py
```diff
@@ -156,17 +141,16 @@ def log(self, mqtt_topic: str, message: str) -> None:
         message : str
             Corresponding message to be recorded
         """
-        time_delta = time.monotonic() - self._START_TIME
+        time_delta = time.monotonic() + self._LAST_RECORDED_TIME
```
kdav108, line 146: Just to confirm, if we use monotonic time, we can't tell from the logs what time the messages logged were actually sent/received right?

DAS/das/V3_mqtt_playback.py
```diff
@@ -9,10 +10,6 @@
     formatter_class=argparse.ArgumentDefaultsHelpFormatter,
 )
 
-parser.add_argument(
-    "filepath", action="store", type=str, help="""Filepath of the csv log"""
-)
-
```
kdav108, line 15: Any reason why we're not taking sql file path as argument?

DAS/README.md
```diff
@@ -9,7 +9,7 @@ These steps assume that you are in the data-acquisition-system root directory an
 1. `cd DAS`
 2. `poetry install`
 3. `poetry shell`
-4. Run any DAS command, _e.g._ `python -m das.V3_mqtt_recorder`
+4. Run any DAS command, _e.g._ `python -m das.V3_mqtt_recorder -v`
```
kdav108, line 12: what does the `-m` stand for here?

kdav108 (changes requested):
Great work on this, SQLite is a nice addition (that's why I opted to jump in and review without a review request 😂).

We still have some old tests in `DAS/das/tests` that needs updating to test sql and not csv output.

Also, what's the difference between `mqtt_wireless_logger.py` and `V3_mqtt_recorder.py` they both seem to log v3 WM topic data.

DAS/README.md
Blake-Haydon: -m stands for module. It allows us to run `das.V3_mqtt_recorder` which is inside a folder. The fancy definition is that it allows us to _execute python code from the command line via modulename rather than filename_.

DAS/README.md
Blake-Haydon: That is definitely a good point. I 100% did not consider that the username and password would be stored in the history.

I think using files is the way to go, however there should always be the option to use a username and password from the command line for quick tests. We could even warn users to use the file option as shown below.

``` 
ALERT: --username AND -password SHOULD ONLY BE USED FOR TESTING
ALERT: USE A .env FILE INSTEAD
```

DAS/das/V3_mqtt_playback.py
Blake-Haydon: There should only be a single database file so the user does not need to input it.

Blake-Haydon:
I will update the tests in `DAS/das/tests` in the next PR @kdav108. It was a little sad to see everything fail 😔, but it should be pretty easy to fix.

DAS/das/utils/logger.py
Blake-Haydon: We will only get relative time from monotonic time. This is because we do not have a real time module on the pi. Due to having it run potentially offline.

kdav108 (approved):
Looks great, approving this with a promise that the tests will be updated with high priority next 😉?

Blake-Haydon:
Already started working on the tests to all should be well by next PR!

PR #81, opened 2021-09-18
Add WM status LED

hallgchris:
## Description

This PR adds code to control the WM's status LED. The status LED basically indicates what state the WM is in, e.g. connecting to MQTT broker or publishing data.

The status LED can be any of the 8 colour combinations made by having each of red, green and blue on or off. We could use PWM to allow many more colours but we probably don't want to have colours that are too similar, so this is fine for now (i.e. less work for me). The LED can also be blinking or solid. Take care that the LED cannot blink during blocking operations, I've made sure that this doesn't occur in the current implementation.

You can see the pre-defined LED states in `status_led.py`. I'll make a guide on notion for using the back WM shortly, which will include this information in a more reader-friendly manner.

Also changed in this PR:
- Battery and data publishing loops are now in their own `uasyncio` tasks, so that they run fully independently and don't block other tasks (e.g. LED blinking) from running when they're idle.
- The WM will now repeatedly try to connect to MQTT until it is unsuccessful. Previously it would just give up after one unsuccessful attempt.

## Screenshots

No screenshots, but here's the LED
image

## Steps to Test

If you happen to have the WM somehow, load the code with
```bash
cd wireless_modules/back_module
./upload.sh /dev/ttyUSB0 # substitute correct device
```

then reboot the WM. You should observe statuses as in `status_led.py`.

### DIFF
--- a/wireless_modules/boot.py
+++ b/wireless_modules/boot.py
@@ -1,8 +1,9 @@
-
 import esp
 import network
-import config
 from machine import Pin
+
+import config
+from status_led import StatusLed, WmState
 
 
 def do_connect(essid, password):
@@ -21,16 +22,19 @@ def do_connect(essid, password):
     station_interface.connect(essid, password)
 
     # Wait for the module to connect
-    while not station_interface.isconnected(): 
+    while not station_interface.isconnected():
         pass
-    
+
     print("Connected, network config:", station_interface.ifconfig())
 
+
+status_led = StatusLed(config.STATUS_LED_PINS)
+status_led.set_state(WmState.ConnectingToNetwork)
 
 # Allows the code to be uploaded with tools such as rshell / ampy
 esp.osdebug(None)
 do_connect(config.ESSID, config.PASSWORD)
 
 # Turn on the built in LED when booted and connected to WiFi (blue light)
-builtin_LED = Pin(config.led_pin, Pin.OUT)
-builtin_LED.on()
+builtin_led = Pin(config.BUILTIN_LED_PIN, Pin.OUT)
+builtin_led.on()
--- a/wireless_modules/mqtt_client.py
+++ b/wireless_modules/mqtt_client.py
@@ -1,4 +1,4 @@
-from umqtt.simple import MQTTClient
+from umqtt.simple import MQTTException, MQTTClient
 import uasyncio as asyncio
 
 
@@ -10,8 +10,9 @@ class Client:
         :param broker_address: A string holding domain name or IP address of the broker to connect to, to send and
                             receive data.
         """
-        self.client = MQTTClient(client_id, broker_address)
+        self.client = MQTTClient(client_id, broker_address, keepalive=60)
         self.mqtt_broker = broker_address
+        self.connected = False
 
     async def connect_and_subscribe(self, topics_to_subscribe, callback_func):
         """
@@ -32,8 +33,8 @@ class Client:
                 )
                 self.client.connect()
                 self.connected = True
-            except OSError as e:
-                print("Failed to connect! {}".format(e))
+            except (OSError, MQTTException) as e:
+                print("Failed to connect! {}: {}".format(type(e), e))
                 print("Reattempting in 5 seconds.")
                 await asyncio.sleep(5)
 
--- a/wireless_modules/mqtt_client.py
+++ b/wireless_modules/mqtt_client.py
@@ -12,6 +12,7 @@ class Client:
         """
         self.client = MQTTClient(client_id, broker_address)
         self.mqtt_broker = broker_address
+        self.connected = False
 
     async def connect_and_subscribe(self, topics_to_subscribe, callback_func):
         """
@@ -40,6 +41,7 @@ class Client:
                 await asyncio.sleep(5)
 
         print("Connected to {}".format(self.mqtt_broker))
+        self.connected = True
 
         # Subscribe to each topic
         for topic in topics_to_subscribe:
@@ -86,4 +88,5 @@ class Client:
         """
         Disconnect from the broker
         """
+        self.connected = False
         self.client.disconnect()
--- a/wireless_modules/mqtt_client.py
+++ b/wireless_modules/mqtt_client.py
@@ -1,4 +1,5 @@
 from umqtt.simple import MQTTClient
+import uasyncio as asyncio
 
 
 class Client:
@@ -12,7 +13,7 @@ class Client:
         self.client = MQTTClient(client_id, broker_address)
         self.mqtt_broker = broker_address
 
-    def connect_and_subscribe(self, topics_to_subscribe, callback_func):
+    async def connect_and_subscribe(self, topics_to_subscribe, callback_func):
         """
         Connects to the MQTT broker and subscribes to each topic in 'topics_to_subscribe' using QoS = 1
         :param topics_to_subscribe: An array of topics to subscribe to.
@@ -22,7 +23,22 @@ class Client:
         self.client.set_callback(callback_func)
 
         # Connect to MQTT broker
-        self.client.connect()
+        connection_success = False
+        while not connection_success:
+            connection_success = True
+            try:
+                print(
+                    "Attempting connection to MQTT broker at {}...".format(
+                        self.mqtt_broker
+                    )
+                )
+                self.client.connect()
+            except OSError as e:
+                connection_success = False
+                print("Failed to connect! {}".format(e))
+                print("Reattempting in 5 seconds.")
+                await asyncio.sleep(5)
+
         print("Connected to {}".format(self.mqtt_broker))
 
         # Subscribe to each topic
--- a/wireless_modules/mqtt_client.py
+++ b/wireless_modules/mqtt_client.py
@@ -10,7 +10,8 @@ class Client:
         :param broker_address: A string holding domain name or IP address of the broker to connect to, to send and
                             receive data.
         """
-        self.client = MQTTClient(client_id, broker_address, keepalive=60)
+        self.keep_alive_interval = 60
+        self.client = MQTTClient(client_id, broker_address, keepalive=self.keep_alive_interval)
         self.mqtt_broker = broker_address
         self.connected = False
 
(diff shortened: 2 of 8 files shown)


wireless_modules/boot.py
```diff
@@ -1,9 +1,10 @@
-
 import esp
 import network
-import config
 from machine import Pin
 
+import config
+from status_led import StatusLed, WmState
```
kdav108, line 6: Nit: Would be nice to group the import statements, seeing as we do that in this repo

wireless_modules/mqtt_client.py
```diff
@@ -22,8 +24,24 @@ def connect_and_subscribe(self, topics_to_subscribe, callback_func):
         self.client.set_callback(callback_func)
 
         # Connect to MQTT broker
-        self.client.connect()
+        connection_success = False
+        while not connection_success:
+            connection_success = True
```
kdav108: Bit pedantic, but since we already have the class variable `self.connected`, we can probably just use that here.

wireless_modules/status_led.py
```diff
@@ -0,0 +1,82 @@
...
+                self.__set_leds_on(self.warning_state)
+                await asyncio.sleep_ms(200)
+
+                # Need to check that self.warning_state is still not None
+                # because it may have been changed during the sleep
+                if self.warning_state and self.warning_state.blink:
+                    self.__set_leds_off()
+                    await asyncio.sleep_ms(200)
+
+            else:
```
kdav108, line 76: What about making this more explicit:`elif self.state` so if we ever call this function without setting any state, the WM won't crash. A very rare scenario, but good to make it crash proof I guess.

wireless_modules/wireless_module.py
```diff
@@ -73,6 +75,26 @@ def sub_cb(self, topic, msg):
...
+            return
+
+        while True:
+            battery_voltage = self.battery.read()
+            if self.mqtt.connected:
+                self.mqtt.publish(
+                    self.battery_topic, ujson.dumps(battery_voltage), retain=True
+                )
+            self.status_led.set_warning_state(
+                WmState.LowBattery if battery_voltage["voltage"] <= 3.3 else None
```
kdav108: Can we make the low battery threshold (`3.3`) into a variable

kdav108 (changes requested):
Great design! Using `ayncio` is quite neat!

Unfortunately couldn't test it physically, but took some extra time to simulate it mentally, and it works well :)

kdav108:
Also, if you get the notion manual written up for the back WM explaining the status LED messages before merging this PR, it might be good to reference it from the README.

hallgchris:
Note to self before I merge, the magenta for the sensor initialising state is too close to blue. Need to try another colour. Also, the LED doesn't blink despite being told to because the initialisation code is blocking, should make it explicitly not blinking.

wireless_modules/boot.py
hallgchris: Aren't thney grouped? They're split into micropython modules and our own files

wireless_modules/status_led.py
hallgchris: Actually, that's why I had the `Undefined` LedState... Just forgot to use it!

hallgchris: Nevermind, I did use it... This state will never happen, unless we explicitly try to set the state to `None`, in which case we want to know about it!

wireless_modules/boot.py
kdav108: I believe we're grouping them on the style, all the `import` statements are together, and the `from ... import ...` are grouped together

wireless_modules/status_led.py
kdav108: It can happen if we initialise `StatusLed` and do not call `set_state` before calling `start_blink_loop`. Although, that case is an error as well, so I'm fine if we leave it as is

kdav108 (approved):
LGTM!

wireless_modules/status_led.py
hallgchris: `set_state` is called by `StatusLed.__init__` :)

wireless_modules/boot.py
hallgchris: I think that's just a coincidence that that's also been the case. My intention has always been to group first python built in modules, then 3rd party modules, then our own. This is what we've done in other repos (where we remember to) also.

hallgchris:
Made some changes. I accidentally caused an issue with MQTT connection in the last review round, which made me realise some types of errors aren't handled properly. Also discovered that the latest mosquitto version doesn't play nicely with the `umqtt` modules default settings: see https://github.com/micropython/micropython-lib/issues/445

Finally, also ran into a bug which I believe is introduced in this PR (come to think of it, haven't checked `master`). Leaving the sensor idling for a few minutes causes a funky exception to occur. I can't quite track down the cause, I wonder if it's to do with `umqtt` not being ideal for async code in the first place (source: https://forum.micropython.org/viewtopic.php?t=9581). Might see if using `mqtt_as` as suggested is any different. Error is below, if you have ideas make them heard:

```
Task exception wasn't retrieved
Traceback (most recent call last):
  File "uasyncio/core.py", line 1, in run_until_complete
  File "wireless_module.py", line 149, in start_data_loop
  File "wireless_module.py", line 127, in wait_for_start
  File "mqtt_client.py", line 77, in check_for_message
  File "umqtt/simple.py", line 204, in check_msg
  File "umqtt/simple.py", line 173, in wait_msg
OSError: -1
```

I haven't been able to find out what `OSError: -1` means.

Otherwise if all else is fine I'd like to get this in and will solve this bug in another PR since it might require a substantial bit of rewriting if we switch MQTT module.

wireless_modules/status_led.py
kdav108: Ahhh don't know how I missed that

hallgchris:
So turns out the problem was the broker was disconnecting us because `umqtt` doesn't respond to the server's pings, nor does it automatically send its own. This wasn't an issue previously because we were configured not to send pings at all, but with mosquitto 2.0.12 this is no longer an allowable default.

The solution was relatively simple in the end. `umqtt` has a method to manually ping the server, so calling that once a minute works as expected.

wireless_modules/wireless_module.py
```diff
@@ -38,6 +42,22 @@ def __init__(self, module_id, battery_reader=None):
...
         self.battery = battery_reader
+        self.status_led = status_led
+
+        asyncio.get_event_loop().set_exception_handler(self.error_handler)
+
+    def error_handler(self, _loop, context):
+        self.status_led.set_state(WmState.Error)
+        print("An error occured in a uasyncio task!")
+        print(context["message"])
+        sys.print_exception(context["exception"])
```
kdav108, line 53: Do we need both these print statements?

wireless_modules/wireless_module.py
```diff
@@ -38,6 +42,22 @@ def __init__(self, module_id, battery_reader=None):
...
+
+        # FIXME: A nicer solution would be to restart all running tasks rather than reset
```
kdav108, line 55: Is there an easy way to gracefully stop all tasks before reseting the esp32?

wireless_modules/wireless_module.py
```diff
@@ -38,6 +42,22 @@ def __init__(self, module_id, battery_reader=None):
...
+
+        print("Resetting in 5 seconds...")
+
+        time.sleep(5)
+        machine.reset()
```
kdav108, line 60: We already catch exceptions and do this in `back_module/main.py`, if errors thrown by async functions reach that `try and catch` block, we can just rely on that to reset the machine.

wireless_modules/wireless_module.py
```diff
@@ -2,22 +2,26 @@
...
 
 from mqtt_client import Client
+from status_led import WmState
 
 try:
     import config
 except FileNotFoundError:
     print("Error importing config.py, ensure a local version of config.py exists")
 
+LOW_BATTERY_THRESHOLD = 3.3
```
kdav108, line 16: It's probably not a concern for the short-medium term, but is it possible in the future for different WM to use different types of batteries and hence have different low battery thresholds?

kdav108 (changes requested):
Thanks for fixing these issues (and picking them up int he first place 😂)! 

Just a few suggestions, may not even need to entertain all of them...

wireless_modules/wireless_module.py
hallgchris: Yep, so by memory the first one will print whatever message is associated with the exception (e.g. `Task exception wasn't retrieved`, looking at the example above) and then the next line prints the actual traceback.

wireless_modules/wireless_module.py
hallgchris: Ordinarily there would be, but asyncio doesn't let you "cancel" a task from within the task itself. This function is considered by asyncio to be part of the task that the error occurred in, so unfortunately there's not a simple clean way to cancel the tasks.

To be honest, there's no harm in resetting the esp32 without notice anyway, unlike for example a raspberry pi which mightn't be as big of a fan. When we turn off the esp32 we don't gracefully stop tasks either, we just cut power.

The main reason it would be good to restart tasks rather than reset would be that we could come back online faster and maintain state such as whether we should be logging. However, if it's something we decide we want I think it could go in its own PR, kinda unrelated to the status LED at this point. Another solution to the problem would be to just make the `start`/`stop` topics retained (I reckon we should)

wireless_modules/wireless_module.py
hallgchris: Good catch, but actually asyncio catches and suppresses the errors, even though it may kill the tasks. It does print the traceback by default, but won't trigger a reset meaning the WM is out of action until manually power cycled.

wireless_modules/wireless_module.py
hallgchris: Very very unlikely. All lithium chemistry batteries have a 3.7 volt nominal voltage and it'd be fairly safe to say that 3.3 will always be a good threshold (or if we decide it's too low or too high, we'd want to adjust it for all WMs). Only possibility of a different voltage is if we use multiple cells, but currently we're looking at significantly reduced power consumption in the front WM, not increased. I think this is fine.

hallgchris:
No changes made lol, easiest review response ever. Let me know your thoughts

wireless_modules/wireless_module.py
kdav108: I thought tracebacks themselves print out the actual message as well

wireless_modules/wireless_module.py
kdav108: So does that mean the try catch block is redundant in the `main.py` script?

wireless_modules/wireless_module.py
kdav108: Yep agree `start/stop` should be retained and completely fine with the reset task to be a seperate PR. I didn't see any benefit in gracefully stopping either, it just seemed nice 😂. Hence my initial emphasise on whether `there is an easy way` to do it.

kdav108 (approved):
Happy days!

wireless_modules/wireless_module.py
hallgchris: The try catch there is still useful in case the initialisation crashes, that won't be caught by asyncio.

PR #82, opened 2021-10-09
Make WM publish to status topic

kdav108:


hallgchris (approved):
@kdav108 I've made a quick couple additions:
- The WM wasn't doing the keep-alive thing correctly, so I quickly fixed that up. Necessary so that the last will works expectedly
- Moved the last will setting into a function, it just means we can avoid the messy checking if both parameters are set. I think it's slightly nicer. Let me know what you think

Kudos for writing code that worked without being able to run it!

Go ahead and merge if you're happy.

kdav108:
I think the changes look clean, thanks Chris!

PR #83, opened 2022-01-31
Add ANT+ service

hallgchris:
## Description

This PR adds a systemd service for the ANT+ script, so that it can easily be run on the bike.

I also updated some READMEs and moved the services to a central folder.

## Screenshots

image

## Steps to Test

Follow the steps in `services/README.md` and verify that *both* services can start correctly. (You'll need to make sure that both the logger and ANT+ script have their dependencies installed)


services/README.md
```diff
@@ -0,0 +1,47 @@
...
+```
+systemctl --user status logger.service
+```
+
+Tip: Add the `-f` flag to continuously monitor new logs
+
+## Requirements
+
+For poetry to correctly clone the `common/mhp` repository using SSH, the public key of the Pi must me shared with our
+github account
```
kdav108: maybe just add that it should be added to `mhp-deployment` GitHub user just to avoid any confusion.

services/ant-plus.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=MHP ANT+ mock wireless module
+
+[Install]
+WantedBy=default.target
+
+[Service]
+Restart=on-failure
+ExecStart=/bin/node ant_plus_logger
```
kdav108, line 9: which node version does `ant_plus_logger` require? I believe `/bin/node` would point to the node installed by root user which on the primary pi is version 10. pi user has node version 14. If we need 14, we can possibly use `sudo su - pi && /bin/node ant...`

hallgchris: Looks like it runs under node 10 - If you're okay with this I might leave it as is. (your solution would still require the user to have manually installed node 14)
image

kdav108: I'm fine if the script works okay with node 10.

kdav108 (approved):

PR #84, opened 2022-02-02
Update logger to use CSV_DIR path for env file

kdav108:
## Description

This PR updates `mqtt_wireless_logger.py` to find the `CSV_DIR` path in a `.env` file if provided. The `CSV_DIR` is where the csv files containing WM data are stored.

Also fixed a bug where the code crashed when the WM is sent a start and stop message but the WM didn't publish anything (possible when the module is not powered up).

## Screenshots

N/A

## Steps to Test

1) Add `CSV_DIR` variable in `.env` file and run `python mqtt_wireless_logger.py` in poetry shell to check that the script is storing data in the provided path.
2) Use `das/V3_fake_module.py` to run fake data for the logger to store.

### DIFF
--- a/DAS/.env.example
+++ b/DAS/.env.example
@@ -1,0 +1,1 @@
+CSV_DIR="~/Documents/dashboard/server/data"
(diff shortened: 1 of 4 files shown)


hallgchris (approved):
LGTM, but add a `.env.example` and put `.env` int the `.gitignore`

PR #88, opened 2022-02-11
Get ANT+ working

hallgchris:
## Description

Kind of a misc. PR
- Document the permission requirements to get the ant plus sensor working
- Update the MQTT message format to not include module ID or null fields (i.e. that sensor has not connected) so that the dashboard type validation passes

## Screenshots

n/a

## Steps to Test

Get an ANT+ dongle and a raspberry Pi, and after following the instructions in the README, confirm that the script can connect to the ANT+ dongle.

### DIFF
--- a/ant_plus_sensor/README.md
+++ b/ant_plus_sensor/README.md
@@ -5,16 +5,29 @@
 ## **[Ant-Plus Logger.js](https://github.com/monash-human-power/data-acquisition-system/blob/master/ant_plus_sensor/ant_plus_logger.js)**
 `ant_plus_logger.js` is the only way we can communicate to ant+ devices. It serves as the middle-man between the ANT+ sensors and the DAS MQTT broker. It allows the sensors to show up as a virtual wireless sensor module.
 
-## **Getting Started**
+## Getting Started
 
 ### Setting up node.js libraries
 To install the necessary libraries for node.js you may need to manually install them instead of running `npm install`.
 
 Run the script with `node ant_plus_logger.js`.
 
-To start this script automatically, see [the services README.md](/services/README.md).
+Sometimes, root permissions are required to connect to the USB dongle, so `sudo` may be required.
 
-## **Legacy Code**
+## Deployment
+
+To start this script automatically, follow the instructions in [the services README.md](/services/README.md).
+
+You will also need to ensure the default `pi` user has permissions to access the ANT+ dongle without `sudo`. Create the following file in `/etc/udev/rules.d/50-antplus.rules` and reboot:
+
+```
+# Allow access to ANT+ USB without sudo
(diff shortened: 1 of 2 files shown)


ant_plus_sensor/README.md
```diff
@@ -5,16 +5,29 @@
...
+
+To start this script automatically, follow the instructions in [the services README.md](/services/README.md).
+
+You will also need to ensure the default `pi` user has permissions to access the ANT+ dongle without `sudo`. Create the following file in `/etc/udev/rules.d/50-antplus.rules` and reboot:
+
+```
+# Allow access to ANT+ USB without sudo
+# https://www.xmodulo.com/change-usb-device-permission-linux.html
+SUBSYSTEMS=="usb", ATTRS{idVendor}=="0fcf", ATTRS{idProduct}=="1008", GROUP="users", MODE="0666"
+SUBSYSTEMS=="usb", ATTRS{idVendor}=="0fcf", ATTRS{idProduct}=="1009", GROUP="users", MODE="0666"
```
kdav108, line 27: Would any of this need to change if we buy a new Ant plus dongle? Just wondering if any of these details are specific to the dongle or common across them all.

ant_plus_sensor/ant_plus_logger.js
```diff
@@ -160,20 +160,10 @@ async function heartRateConnect(antPlus) {
...
-            type: 'cadence',
-            value: cadence,
-          },
-          {
-            type: 'heartRate',
-            value: heartRate,
-          },
+          ...(power ? [{ type: 'power', value: power }] : []),
+          ...(cadence ? [{ type: 'cadence', value: cadence }] : []),
+          ...(heartRate ? [{ type: 'heartRate', value: heartRate }] : []),
```
kdav108, line 166: Does dashboard handle an empty array for all these types?

kdav108 (approved):
Nice and small PR, but thanks for all the hard work that went behind debugging and getting to this solution.

I just have 2 questions, only because I don’t have a dongle to test it with myself :)

ant_plus_sensor/README.md
hallgchris: Nope, not unless a new version is released. `0x1008` is for version 2 of the garmin ant+ dongle, and `0x1009` for version 3. Right now they (and clones) are the only two dongles supported by the ant plus library.

ant_plus_sensor/ant_plus_logger.js
hallgchris: Because of the `...` these arrays are being destructured into the parent array. So, if these fields are null, their entry won't be included in `sensors` array at all.

Think of it as either appending the array `[{ type: 'power', value: power }]` if `power` is not null, else `[]` to the sensor array.

PR #90, opened 2022-03-09
Make ANT plus send message on status topic

kdav108:
## Description

Currently the ANT+, our 4th WM, did not send any indication of whether it's online or not. This PR makes it so the script sends a message indicating it's online when it has connected to the MQTT client and has detected the ANT+ stick. And the script will send an offline message when it disconnects from the broker.

Note: I have made it send the online message after the ANT+ stick is initialised so that the meaning of 'online' for any WM is that it's ready to send data (which is its importance) not that it's just connected to the broker.

## Screenshots
Note: there's no sensor data since I have not attached the other peripherals needed.
Screen Shot 2022-03-09 at 5 35 50 pm

## Steps to Test

1) Run the ant plus script, `node ant_plus_logger.js`
2) Run dashboard
3) Check that the ANT+ status is online
4) To test that the messages are retained correctly, kill the dashboard server and re-run and check that the ANT+ status is still online.
5) Exit the ant plus script and confirm the status on dashboard has turned to offline
6) To test the retain message of the will payload, you can again restart dashboard server and confirm the status remains offline.

### DIFF
--- a/ant_plus_sensor/ant_plus_logger.js
+++ b/ant_plus_sensor/ant_plus_logger.js
@@ -38,6 +38,7 @@ winston.info(`Wireless module ID: ${moduleID}`);
 const startTopic = `/v3/wireless_module/${moduleID}/start`;
 const stopTopic = `/v3/wireless_module/${moduleID}/stop`;
 const dataTopic = `/v3/wireless_module/${moduleID}/data`;
+const statusTopic = `/v3/wireless_module/${moduleID}/status`;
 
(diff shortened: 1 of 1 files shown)


hallgchris (commented):
Sorry, I missed the review request when you made this.

LGTM, but what do you think about the case when the module is switched on but no dongle is connected? In this case, nothing is published at all.

hallgchris (approved):

PR #91, opened 2022-03-30
install mqtt-logger package

Blake-Haydon:
## Description

This PR installs `mqtt-logger` from PyPi instead of using a local version. This is meant to simplify testing and development of the logger.

## Steps to Test

Check that `poetry run python -m das.V3_mqtt_recorder /v3/#  -v --host broker.hivemq.com` runs.


hallgchris (commented):
Code edits look good to me, what's blocking this PR?

Blake-Haydon:
@hallgchris There were changes to `mqtt-logger` that fixed a few bugs. I have also tested it with the recorder and playback commands and they seem to work!

DAS/pyproject.toml
```diff
@@ -5,13 +5,14 @@ description = ""
 authors = ["Blake <bhay0010@student.monash.edu>"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.9"
```
kdav108, line 8: Just to confirm, will the repo not work with python 3.7 going forward?

Blake-Haydon: The `DAS` folder and all the files contained in it will not work with python 3.7 anymore. I was running into issues with python 3.7 with `mqtt-logger`. Is there any reason that we need to stay on python 3.7?

kdav108: it's just a pin running` poetry install` on the Pi 😅, if we change python version we would need to run poetry install again in the new virtual env. But it should be alright, as long as we don't pull these changes onto the raspberry pi.

Blake-Haydon: Ah that makes sense. I'll try and decrease the python version requirements.

Blake-Haydon:
Will pause this PR for now as I work on testing the excel creation tool for `mqtt-logger` captured data.

Blake-Haydon:
This PR is now in a good state. Will test functionality this Saturday.

kdav108 (approved):
LGTM!

PR #98, opened 2022-04-07
Add ant+ speed sensor

hallgchris:
## Description

Added the ant+ speed sensor to the existing ant+ dongle WM. Also, disabled the reed switch sensor on the back WM.

## Screenshots

n/a

## Steps to Test

Tested and working on the hardware. Code review is sufficient :)

### DIFF
--- a/ant_plus_sensor/ant_plus_logger.js
+++ b/ant_plus_sensor/ant_plus_logger.js
@@ -153,6 +153,9 @@ async function heartRateConnect(antPlus) {
 (async () => {
   let isRecording = false;
   let speed = 0;
+  // The number of wheel revolutions according to the sensor when we start recording
+  let startWheelRevolutions = null;
+  let distance = 0;
   const powerAverage = new RollingAverage(3000);
   let cadence = 0;
   let heartRate = 0;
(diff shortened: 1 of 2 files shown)


ant_plus_sensor/ant_plus_logger.js
```diff
@@ -156,6 +187,21 @@ async function heartRateConnect(antPlus) {
     }
   });
 
+  const bicycleSpeedScanner = await bicycleSpeedConnect(antPlus);
+  bicycleSpeedScanner.on('speedData', (data) => {
+    // Store speed into global variable
+    speed = data.CalculatedSpeed;
+    if (startWheelRevolutions === null) {
+      startWheelRevolutions = data.CumulativeSpeedRevolutionCount;
```
kdav108, line 195: Just for my own understanding, startWheelRevolutions would ideally be 0 right?

kdav108 (approved):
LGTM! Good to see it didn’t take much code changes to add and remove a sensor.

ant_plus_sensor/ant_plus_logger.js
hallgchris: Yeah so I think `CumulativeSpeedRevolutionCount` is actually stored on the ant+ sensor and sent to the dongle with each message. This means that if the speed sensor is turned on (which it will do automatically when it detects wheel movement) it will start counting. This way we can set an appropriate offset.

PR #100, opened 2022-04-13
fix: not excel tab with no data

Blake-Haydon:
## Description

Does not produce an excel tab if no data has been logged.

## Steps to Test

1. Run [SQL Logger](https://github.com/monash-human-power/data-acquisition-system/tree/master/DAS#v3-mqtt-recorder) with one or more wireless modules off
2. Run [log_2_excel.py](https://github.com/monash-human-power/data-acquisition-system/blob/master/DAS/das/log_2_excel.py)
3. Check that the generated excel file does not have any empty sheets.


kdav108 (approved):
Sorry for taking so long, just slipped off my mind.

hallgchris:
Didn't even know this existed haha. I don't think my brain was fully functional the week before ford

PR #103, opened 2022-08-23
Implement V3 start topic for WM and ant+ logger

AroshHeenkenda:
## Description
Implementing the new V3 start topic, **'v3/start**, to coincide with the V3 software system. This will be replacing the old start and stop topics. The new topic will have a structure of **{"start": true/false}**, where the Boolean value will denote whether any logging needs to start or not. There are 2 different components that have been altered the **Wireless Module** and **ANT+ Logger**, as such there are 2 different systems to test. 

## Screenshots
**ANT+ Screenshots**
Zadig Install:
zadig install

ANT+ Failed Connection:
Inkedfailed connection

ANT+ Successful Connection:
Inkedsucessful connection

V3 Start True:
Inkedv3start true

V3 Start False:
Inkedv3start false

**Wireless Module Screenshots**
Putty Config:
putty

On Start:
on start

WM True:
v3 start true small

WM False:

## Steps to Test
**Testing ANT+ Changes**
1. Switch to `.\ant_plus_sensor` (in a terminal).
2. Run 'npm install' to install necessary dependencies if you haven't already.
3. Ensure an ANT+ USB dongle is connected to your computer.
4. Windows users may need to install [Zadig](https://zadig.akeo.ie/) to read the dongle correctly. 
   - Install the latest version of Zadig.
   - Follow the USB installation steps.
   - See _Zadig Install,_ ensure that **libusb-win32**  is selected.

5.  In the terminal write `node ant_plus_logger.js`.
    - If code fails to find ant-plus USB and is stuck as in _ANT+ Failed Connection_.
    - May need to change instances of `GarminStick3` to `GarminStick2`.
    - These are found on lines 82, 88, 92, 106, 123, 139 (in `ant_plus_logger.js`).
    - A successful connection will look like _ANT+ Successful Connection_.
6.  Enure MQTT broker is started, then publish V3 start topics.
    - `mosquitto_pub -t 'v3/start' -m '{\"start\":true}'`, output should look like _V3 Start True_.
    - `mosquitto_pub -t 'v3/start' -m '{\"start\":false}'`, output should look like _V3 Start False_.
    - Note that Mac users don't need to use **\\"** to ensure the string literal is printed correctly.

**Testing Wireless Module Changes**
1. Connect the wireless module (WM) to your computer via micro-USB.
2. Check which port it is connected to and note it
    - Windows users can use **Device Manager** to check this.
3. Turn on the primary display.
4. Set up a mobile hotspot with the name `MHP_MobileNet` with the normal password (found on Notion).
    - Ensure your device also connects to this network.
5.  Flash the wireless module code using:
     -   `ampy -p port_WM_is_connected_to -b 115200 put path_to_wireless_module.py`.
     - Note that [ampy](https://pypi.org/project/adafruit-ampy/) does need to be installed for this, however any other means of flashing to WM is fine.
6. Once flashing is complete disconnect and then reconnect the micro-usb cable.  
7. Use [Putty](https://www.putty.org/) or similar software to analyze the output of the WM.
    - See _Putty Config_ for the configuration, note that you should use the port that you noted earlier.
    - On connect you should see a message similar to _On Start_.
8. Now publish the V3 start topics.
    - `mosquitto_pub -t 'v3/start' -m '{\"start\":true}' -h mhp-v3-primary.local`, the output should look like _WM True_.
    - `mosquitto_pub -t 'v3/start' -m '{\"start\":false}' -h mhp-v3-primary.local`, the output should look like _WM False_.
9. Once testing is over ensure that you flash the `master` version of `wireless_module.py`, using the same terminal command from earlier.    

**Notes**
- Sorry for the long PR, just need to make sure that nothing goes wrong with testing.
- Also sorry to Mac and Linux users, I'm not completely sure about what you'd need to use for testing.

### DIFF
--- a/ant_plus_sensor/ant_plus_logger.js
+++ b/ant_plus_sensor/ant_plus_logger.js
@@ -40,8 +40,6 @@ const args = argumentParser.parseArgs();
 const { id: moduleID, host: mqttAddress, rate } = args;
 winston.info(`Wireless module ID: ${moduleID}`);
 
-// const startTopic = `/v3/wireless_module/${moduleID}/start`;
-// const stopTopic = `/v3/wireless_module/${moduleID}/stop`;
 const v3StartTopic = 'v3/start';
 const dataTopic = `/v3/wireless_module/${moduleID}/data`;
 const statusTopic = `/v3/wireless_module/${moduleID}/status`;
@@ -168,25 +166,6 @@ async function heartRateConnect(antPlus) {
   // Announce we're online once ANT+ stick is also connected
   mqttClient.publish(statusTopic, JSON.stringify(onlineMsg), { retain: true });
 
-  // mqttClient.subscribe([startTopic, stopTopic]);
-  // mqttClient.on('message', (topic) => {
-  //   winston.info(`Topic fired: ${topic}`);
-  //   switch (topic) {
-  //     case startTopic:
-  //       isRecording = true;
-  //       distance = 0;
-  //       startWheelRevolutions = null;
-  //       winston.info('Start publishing data');
-  //       break;
-  //     case stopTopic:
-  //       isRecording = false;
-  //       winston.info('Stop publishing data');
-  //       break;
-  //     default:
-  //       winston.error(`Unexpected topic: ${topic}`);
-  //       break;
-  //   }
-  // });
   mqttClient.subscribe(v3StartTopic);
   mqttClient.on('message', (topic, payload) => {
     winston.info(`Topic fired: ${topic}`);
--- a/ant_plus_sensor/ant_plus_logger.js
+++ b/ant_plus_sensor/ant_plus_logger.js
@@ -40,8 +40,9 @@ const args = argumentParser.parseArgs();
 const { id: moduleID, host: mqttAddress, rate } = args;
 winston.info(`Wireless module ID: ${moduleID}`);
 
-const startTopic = `/v3/wireless_module/${moduleID}/start`;
-const stopTopic = `/v3/wireless_module/${moduleID}/stop`;
+// const startTopic = `/v3/wireless_module/${moduleID}/start`;
+// const stopTopic = `/v3/wireless_module/${moduleID}/stop`;
+const v3StartTopic = 'v3/start';
 const dataTopic = `/v3/wireless_module/${moduleID}/data`;
 const statusTopic = `/v3/wireless_module/${moduleID}/status`;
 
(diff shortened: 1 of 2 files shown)


ant_plus_sensor/ant_plus_logger.js
```diff
@@ -167,23 +166,27 @@ async function heartRateConnect(antPlus) {
...
+  mqttClient.on('message', (topic, payload) => {
     winston.info(`Topic fired: ${topic}`);
-    switch (topic) {
-      case startTopic:
+
+    if (topic === v3StartTopic) {
+      const msg = JSON.parse(payload);
+      winston.info(
+        `Received data from topic ${topic}: ${JSON.stringify(msg.start)}.`,
+      );
```
kdav108: Nit: We  already have a log statement on line 171, this one would just be repetitive.

wireless_modules/wireless_module.py
```diff
@@ -138,6 +138,11 @@ async def wait_for_start(self):
             # Start message received, tell sensors to start
             for sensor in self.sensors:
                 sensor.on_start()
+        
+        #Fix for changing LED when a true is retained
+        if self.start_publish and self.status_led.state != WmState.Publishing:
+            self.status_led.set_state(WmState.Publishing)
```
kdav108: Thought of an alternative: we can set the led state in `sub_cb`, so whoever we receive start message we set led in that method itself. WDYT?

kdav108 (changes requested):
Looking good so far, just a few suggestions...

ant_plus_sensor/ant_plus_logger.js
AroshHeenkenda: I think I left that there from when I was testing, as it is redundant I've removed it from the code.

wireless_modules/wireless_module.py
AroshHeenkenda: Yeah, this would be much better. That way if we received a retained topic we would automatically switch the LED rather than checking in the `wait_for_start` function, right?

kdav108: Yeah exactly. We can then remove `self.status_led.set_state(WmState.Publishing)` and `self.status_led.set_state(WmState.Idle)` from `wait_for_start()` and move them to `sub_cb()`

AroshHeenkenda:
@kdav108 when we process the V3 start message if it's _True_ we set `self.status_led.set_state(WmState.Publishing)`, do we also need to set that to `self.status_led.set_state(WmState.Idle)`, if it is _False_?

wireless_modules/wireless_module.py
```diff
@@ -137,7 +143,7 @@ async def wait_for_start(self):
 
             # Start message received, tell sensors to start
             for sensor in self.sensors:
-                sensor.on_start()
+                sensor.on_start()
```
kdav108, line 146: Nit: Can remove this spacing since python is quite strict with whitespaces

kdav108:
> @kdav108 when we process the V3 start message if it's _True_ we set `self.status_led.set_state(WmState.Publishing)`, do we also need to set that to `self.status_led.set_state(WmState.Idle)`, if it is _False_?

Yep that's right

kdav108 (approved):
Looking good!

PR #104, opened 2022-08-24
Create different poetry env for log_2_excel based on py3.7

kdav108:
## Description

Currently, the python project for logger (`DAS/`) requires python 3.9. The dependency is because the mqtt-logger library requires python 3.9. However, raspberry pi does not support `numpy` library for python 3.9, causing the `log_2_excel` script (which requires bumpy) to be useless on the raspberry pi.

However, we need `log_2_excel` to work on the raspberry pi in order to automate uploading excel files of the logs. The solution is to move `log_2_excel` outside the logger project (basically move it outside `DAS/`) and create it's own poetry env using python 3.7. This new 'service'/software component will be where the file upload code can be situated in the future.

## Screenshots

N/A

## Steps to Test

1) Test that you can still run `V3_mqtt_recorder` with the existing poetry environment.
2) Test `log_2_excel` works locally and on raspberry pi with the new poetry environment.


kdav108:
No longer needed since we were able to downgrade python requirement from 3.9 to 3.7 for mqtt-logger

PR #107, opened 2022-09-23
Downgrade python version requirement

kdav108:
## Description

2 small but important changes:
- Downgrade python version requirement for DAS poetry shell
- Updated `V3_fake_module.py` (our script to generate fake WM data for testing) to use our new `v3/start` topics

## Screenshots

N/A

## Steps to Test

1) Delete your current poetry env:
```shell
# find poetry env in /DAS
poetry env list

# delete env
poetry env remove <env-name>
```

2) Run `poetry shell` - this should spawn a new poetry shell with python 3.7, if it doesn't you might need to use `pyenv local 3.7` to set the local python version to 3.7

3) Run `poetry install` and then the logger script and log_2_excel:
```
python V3_mqtt_recorder.py

# For this you may need to set up .env file (see .env.example)
python log_2_excel.py
```

4) Also test running `python V3_fake_module.py` and it should send fake data over the WM data topic
5) In a new terminal subscribe to `v3/start`:
```
mosquito_sub -t 'v3/start' -v
```
6) When you run `V3_fake_module.py`, in your subscribed terminal you should see a message similar to `{'start': true}` and when you end the fake module script u should receive the following message: `{'start': false}`

### DIFF
--- a/DAS/pyproject.toml
+++ b/DAS/pyproject.toml
@@ -5,20 +5,20 @@ description = ""
 authors = ["Blake <bhay0010@student.monash.edu>"]
 
 [tool.poetry.dependencies]
-python = "^3.9"
-paho-mqtt = "^1.5.0"
-urllib3 = "^1.25.9"
-pandas = "^1.0.3"
+python = "^3.7"
+paho-mqtt = "^1.6.1"
+pandas = "1.1.1"
 argparse = "^1.4.0"
-mhp = {git = "git@github.com:monash-human-power/common.git", rev = "202012.9"}
-python-dotenv = "^0.19.2"
-mqtt-logger = "^0.3.1"
-rich = "^12.2.0"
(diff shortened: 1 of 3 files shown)


DAS/pyproject.toml
```diff
@@ -5,20 +5,20 @@ description = ""
 authors = ["Blake <bhay0010@student.monash.edu>"]
 
 [tool.poetry.dependencies]
-python = "^3.9"
-paho-mqtt = "^1.5.0"
-urllib3 = "^1.25.9"
-pandas = "^1.0.3"
```
kdav108, line 11: These have just been re-arranged because I wasn't able to simply downgrade the python version here, I had to delete `pyproject.toml` and recreate it with `poetry init`. Hence, all the libraries are re-ordered.

jmark-9 (approved):
Looks good 👍

PR #114, opened 2023-01-23
Live Anemometer Data

chuiwinnie:
## Description

Created Python scripts for wireless module and sensor base, where new sensors with scripts written in Python can be added to in the future.

Integrated the anemometer (wind sensor) into the data acquisition system (DAS). This allows the anemometer to collect wind speed and direction data, and publish these data to the MQTT broker under the topic `/v3/wireless_module/5/data`.

## Steps to Test
1. Install the [Vaisala USB Instrument Finder driver](https://go.vaisala.com/software/WXT530/Vaisala_WXT530_Configuration_Tool_Weather_Measurement.zip?_ga=2.138439603.1803271655.1674458831-1555859295.1674458831) using these [installation instructions](https://docs.vaisala.com/r/M211840EN-F/en-US/GUID-6D206CCD-21E9-4E9A-98C9-760C90EA90BF/GUID-AE6CDFA9-16A5-4E47-B354-37C04534C558).

2. Connect the anemometer to your computer and note down which port it is connecting to.

3. In the `wind_module` directory, create a local version of `config.py` using the `config.example.py` file. 

4. Get inside the `wind_module` directory using:
     ```
     cd wireless_mdoules_py/wind_module
    ```

5. Ensure that you have [poetry](https://python-poetry.org/) installed, spawn into the poetry environment using:
    ```
    poetry shell
    ```

6. Run the main program using:
    ```
    python main.py
    ```
    Running main program

7. Open 2 terminals:
    - One to subscribe to the topic `/v3/wireless_module/5/data`.
    - Another to publish to the topic `v3/start`.

8. Publish a true start message to the topic `v3/start`. Streams of data should be shown in the terminal subscribed to the topic `/v3/wireless_module/5/data`.

    Publishing data

9. While the data is being published, mess around with the anemometer to ensure values are changing accordingly.

10. Publish a false start message to the topic `v3/start`. The terminal subscribed to the topic `/v3/wireless_module/5/data` should look like this:

    Stop publishing data


wireless_modules_py/sensors/wind_sensor.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Initialise the anemometer to read wind direction and wind speed data.
+        :param port: The serial port that the anemometer is connected to.
+        """
+        self.baudrate = 19200
+        self.sensor = serial.Serial(port, self.baudrate)
+
+        self.readings = []
+
+        # track the last time read() was called
```
AroshHeenkenda: Don't need this comment.

wireless_modules_py/sensors/wind_sensor.py
```diff
@@ -0,0 +1,77 @@
+from sensor_base import Sensor
+import serial
+import time
+
+
+class WindSensor(Sensor):
+    """
+    A class for an anemometer connected via serial port.
```
AroshHeenkenda, line 8: Add more documentation here, include the attributes of this class, this reduces the need for redundant comments in your code.

wireless_modules_py/wind_module/main.py
```diff
@@ -0,0 +1,39 @@
+import sys
+sys.path.insert(1, '../sensors')
```
AroshHeenkenda, line 2: If you can find a better way to do this that would be good, but if not then leave it as is.

AroshHeenkenda (changes requested):
All of your code looks pretty good but you don't need to have so many comments, as your code is readable most people will understand what it does without needing comments. Remove any redundant comments, and try to change how we import modules but if that's the best way leave it as it is.

wireless_modules_py/sensors/wind_sensor.py
chuiwinnie: By attributes of this class, did you mean variables like baudrate, sensor, readings etc.? If so, I'm just wondering if it's necessary to document them in the docstring for the class? Since all these attributes can be easily found inside the init method and their names are pretty self-explanatory?

wireless_modules_py/sensors/wind_sensor.py
chuiwinnie: Will remove this line.
Other than this line, were there any other comments in particular that you found redundant and should be removed?

wireless_modules_py/wind_module/main.py
chuiwinnie: I tried structuring the folders as packages by adding an `__init__.py` in `sensors` and `wireless_module_py` so that we could import like how we normally would but had no luck. As a result, had to use the `sys` module and modify the search path to import `WindSensor` and `WirelessModule`.

wireless_modules_py/sensors/wind_sensor.py
AroshHeenkenda: Yeah you're right, you can disregard this suggestion.

wireless_modules_py/wind_module/main.py
AroshHeenkenda: That's fine continue using `sys`, it works so there's no harm in using it. If we can find a better method to use in the future we can just change the code.

AroshHeenkenda (commented):
As I said before all the code looks good, all you have to do is get rid of any unnecessary comments.

wireless_modules_py/wind_module/main.py
```diff
@@ -0,0 +1,35 @@
...
+
+async def main():
+    logging.basicConfig(
+        format="%(levelname)-8s [%(filename)s] %(message)s", level=logging.DEBUG
+    )
+
+    wind_sensor_port = config.PORT
+    my_wind_sensor = WindSensor(wind_sensor_port)
+
+    # set up the wind module
```
AroshHeenkenda: We know this, no need to comment.

wireless_modules_py/wind_module/main.py
```diff
@@ -0,0 +1,35 @@
...
+    wind_module = WirelessModule(MODULE_NUM)
+    sensors = [my_wind_sensor]
+    wind_module.add_sensors(sensors)
+
+    logging.debug("Start asyncio loop")
+    asyncio.create_task(wind_module.run())
+
+    # keep code running indefinitely
```
AroshHeenkenda: Same deal, we know this no need to comment.

wireless_modules_py/wireless_module.py
```diff
@@ -0,0 +1,182 @@
...
+
+    async def run(self, data_interval=1):
+        """
+        Start running the wireless module. Connect to MQTT and start the data loop.
+        :param data_interval: An integer representing the number of seconds to wait before sending data.
+        """
+        self.mqtt.on_connect = self.on_connect
+        self.mqtt.on_message = self.on_message
+
+        # connect to broker
```
AroshHeenkenda: Redundant comment.

wireless_modules_py/wireless_module.py
```diff
@@ -0,0 +1,182 @@
...
+        self.mqtt.username_pw_set(config.USERNAME, config.PASSWORD)
+        self.mqtt.connect(config.MQTT_BROKER)
+
+        # subscribe to start topic
```
AroshHeenkenda: Another one.

AroshHeenkenda (approved):
Looks good and works good, it should be good to use at AARC.

PR #120, opened 2023-08-10
commented out crash alert logging enabler

da-food:
## Description

Commented out logging setting for crash alert client

## Steps to Test
install on primary then from terminal run (tail -f /var/log/syslog) via ssh


jgOhYeah (approved):
Looks good

PR #124, opened 2023-09-09
Arosh heenkenda/logger alterations

AroshHeenkenda:
## Description

Changing the logging to only make sqlite DB logs, will do manual conversion to excel for duration of BM.


BoneFiend (approved):
Looks good. Try to avoid annoying white space in da future.

PR #130, opened 2025-07-16
Can bus to mqtt

ArnavSingh04:
changed main.cpp and sensor base to connect to wifi and mqtt instead of sending to raspberry via canbus


wired_module/wired_module_arduino/src/main.cpp
```diff
@@ -0,0 +1,95 @@
...
+# include <ICM42670.h>
+
+# include <vector>
+# include "driver/i2c.h"
+# include "DPS368.cpp"
+
+// Wi-Fi and MQTT Settings
+const char* ssid = "A 2 jang pura extn_5G";
+const char* password = "act12345";
+const char* mqttServer = "192.168.0.117";  // will replace it later with raspberry's IP
```
AroshHeenkenda: These shouldn't be written and committed directly into the main file. Instead we should use a `.env` or some similar type of configuration file. The contents of this file in this case will contain something like the below:
``` 
SSID="some_SSID"
PASSWORD="some_password"
MQTT_SERVER="ip_for_server"
```

The `main.cpp` will read from this file, the file should also be `gitignored` so we don't put these secrets onto Git. This way for anyone's test environment they can play around with the values to make it work and it keeps our data confidential. Let me know if you have any questions about this.

AroshHeenkenda (changes requested):
A few things so we can thorougly test this before any merges. 
- I left a comment about using config/.env files, please take a look at that
- Could we please get some steps to test this on our own devices?
- You have a lot of print statements in here which is good, but it would be better if these went to a `.log` file (which would be gitignored), that way we can track what happens on  a device accurately. 

Log file printing should follow a message format like below:
```bash
[yyyy/mm/dd hh:mm:ss] [message_type] log_message

e.g.
[2025/07/17 12:25:00] [INFO] Connected to MQTT service.
```

